Issue #3, opened 2018-10-20
MSBuild integration


adamralph:
Essentially, making this work: https://github.com/adamralph/min-ver#quick-start

- [x] most of the work
- [x] [add the targets file to MSBuildAllProjects](https://github.com/adamralph/min-ver/issues/3#issuecomment-433724484)
- [x] ~[suppress the target for design-time builds](https://github.com/adamralph/min-ver/issues/3#issuecomment-433724484)~ - _moved to https://github.com/adamralph/min-ver/issues/55_

thefringeninja:
Hate to say it but I think you need an msbuild task here. https://github.com/natemcmaster/Yarn.MSBuild might be an inspiration

bording:
@adamralph There are two main ways you can integrate the logic into MSBuild:

1. Ship a command-line tool of some sort that gets called from an MSBuild target via an `Exec` task. 
2. Create an MSBuild task that can be called from a target

Option 1 means you have to output all the needed values to the console and them parse them back out in your MSBuild targets, which can be messy.

Option 2 still means you could also ship a separate command-line tool if you want (for example as a .NET Core global tool), but the actual logic would be invoked directly as a task, and then you can get all the needed values back as regular properties.

Option 2 might seem like the best option, but there is a caveat here as well. Because you're using LibGit2Sharp, getting an MSBuild task working fully cross-platform is non-trivial. There are platform-specific native binaries that you have to deal with. MSBuild offers no assistance in making sure the correct one is loaded, so you have to handle all of that yourself. At the moment, that would involve some level of RID analysis to choose the correct binary. There are LibGit2Sharp changes coming "soon" that will make all of this more palatable, though.

Invoking a command-line tool instead could leverage .NET Core's ability to load the correct binary for you, so that's one reason that might be better. This is likely to be slower though, and this something needs to be invoked every time you build.

thefringeninja:
I think option 2) is the the most viable. Publish a dotnet cli tool e.g. `dotnet min-ver`. Then, using the technique described here: https://natemcmaster.com/blog/2017/07/05/msbuild-task-in-nuget/ you could get that to invoke via `dotnet build`.

Also, I don't think slowness is an issue here... normally you would invoke this as part of your CI process, not on your dev machine.

bording:
[...]

PR #13, opened 2018-10-21
add CLI

adamralph:
Connects to #3


adamralph:
Merging this now so we have a CLI to work with. The rest of the MSBuild integration work can come later.

PR #42, opened 2018-10-28
Targets tweaks

bording:
This make a couple small changes to the targets files:

- Adding the xml namespace means that these targets files can be properly imported if someone tries to use VS 2015. If you don't care about this, or actively want to prevent it, I can remove this commit.
- The GetPackageVersionDependsOn property needs to be set so that when a project reference is used to create a package dependency, the generated nuspec has the correct version for the project reference.


adamralph:
🎉 awesome, thanks!

adamralph:
Released in [alpha 4](https://www.nuget.org/packages/MinVer/1.0.0-alpha.4).

PR #45, opened 2018-10-28
Get version from console output without writing to file

bording:
This removes the hack of writing/reading from a file to get the version from the output.

Instead, each line of the output is checked to see if it starts with `MinVer:`, which all of the log lines will have.

If a line doesn't have that, then it is assumed the line contains the version, and that value will be used in the `MinVerVersion` property.

### DIFF
--- a/MinVer/MinVer.targets
+++ b/MinVer/MinVer.targets
@@ -7,14 +7,15 @@
   </PropertyGroup>
 
   <Target Name="MinVer_GetVersion" Condition=" '$(MinVerVersion)' == '' ">
+    <Exec Command="dotnet &quot;$(MSBuildThisFileDirectory)MinVer.Cli.dll&quot; &quot;$(MSBuildProjectDirectory)&quot;" ConsoleToMSBuild="true">
+      <Output TaskParameter="ConsoleOutput" ItemName="MinVerConsoleOutputItems" />
+    </Exec>
+    <ItemGroup>
+      <MinVerConsoleOutputVersion Include="@(MinVerConsoleOutputItems)" Condition="'$([System.String]::new(`%(Identity)`).StartsWith(`MinVer:`))' != 'true'" />
+    </ItemGroup>
     <PropertyGroup>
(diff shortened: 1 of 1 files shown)


MinVer/MinVer.targets
```diff
@@ -7,14 +7,15 @@
   </PropertyGroup>
 
   <Target Name="MinVer_GetVersion" Condition=" '$(MinVerVersion)' == '' ">
+    <Exec Command="dotnet &quot;$(MSBuildThisFileDirectory)MinVer.Cli.dll&quot; &quot;$(MSBuildProjectDirectory)&quot;" ConsoleToMSBuild="true">
+      <Output TaskParameter="ConsoleOutput" ItemName="MinVerConsoleOutputItems" />
+    </Exec>
+    <ItemGroup>
+      <MinVerConsoleOutputVersion Include="@(MinVerConsoleOutputItems)" Condition="'$([System.String]::new(%(Identity)).StartsWith(`MinVer:`))' != 'true'" />
```
adamralph: Oooh! Nice idea! 💯

adamralph: This needs a tweak though. I hacked the `test-package` target to actually make a commit before building (I'll get the change into `master` ASAP), and I got:

> MinVer.targets(14,73): error MSB4184: The expression "[System.String]::new(MinVer: Using commit    { Sha: 438d6061081fbb103005498cab6096c6b88679ab, Height: 0, Tag: '(none)', Version: 0.0.0-alpha.0 })" cannot be evaluated.

I guess it needs to be wrapped in quotes?

```suggestion
      <MinVerConsoleOutputVersion Include="@(MinVerConsoleOutputItems)" Condition="'$([System.String]::new(&quot;%(Identity)&quot;).StartsWith(`MinVer:`))' != 'true'" />
```

bording:
@adamralph I had to add backticks around the `%(Identity)` bit.

adamralph:
Thanks @bording!

adamralph:
Released in [alpha 5](https://www.nuget.org/packages/MinVer/1.0.0-alpha.5).

PR #51, opened 2018-10-29
isolate try block in Versioner

adamralph:
Connects to #3.


adamralph:
Released in [alpha 6](https://www.nuget.org/packages/MinVer/1.0.0-alpha.6).

bording:
Out of curiosity, how does this connect to #3?

adamralph:
I figured the directory traversal is required for the MSBuild integration, since the targets file can only specify the directory of the current project. I'll grant you that it's only a loose connection. 🙂

PR #53, opened 2018-10-29
Set MSBuildAllProjects property

bording:
This sets the `MSBuildAllProjects` property that is used in MSBuild < 16 to ensure new builds are properly triggered when targets files change.

Connects to #3


adamralph:
Release in [alpha 7](https://www.nuget.org/packages/MinVer/1.0.0-alpha.7).

PR #58, opened 2018-10-31
Don't run targets in design-time builds

### LINKED ISSUE #55: Suppress the MinVer target for design time builds
_Relates to #3._

_From https://github.com/adamralph/min-ver/issues/3#issuecomment-433735623 by @bording:_

> Visual Studio runs a build behind the scenes for various reasons. When it does this, it doesn't actually compile the code, but it does figure out what the inputs to csc would be. These are known as design-time builds. It's important to keep this as minimal as possible to avoid causing problems inside VS. If the targets you're adding don't actually have an impact on the compiler input, then you can condition them to be skipped in a design-time build.
> 
> Since these targets are only changing version properties, I would think it would be safe to skip them.

_From https://github.com/adamralph/min-ver/issues/3#issuecomment-434117809 by @bording:_

> I also look a look at the design-time build stuff, and it turns out that we need to let this run during design-time builds in order to get project references versioned correctly in the generated nuspec. I'm not entirely sure why that is. It feels like it might be a bug, but I need to do a bit more research before filing an issue about it on NuGet/Home.

bording:
Fixes #55

### DIFF
--- a/MinVer/MinVer.targets
+++ b/MinVer/MinVer.targets
@@ -21,7 +21,7 @@
     </PropertyGroup>
   </Target>
 
-  <Target Name="MinVer" BeforeTargets="CoreCompile;GenerateNuspec" DependsOnTargets="MinVer_GetVersion">
+  <Target Name="MinVer" BeforeTargets="CoreCompile;GenerateNuspec" DependsOnTargets="MinVer_GetVersion" Condition="'$(DesignTimeBuild)' != 'true'">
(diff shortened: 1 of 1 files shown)


adamralph:
Thanks @bording! I'll release this in alpha 9 tomorrow.

adamralph:
Release in [alpha 9](https://www.nuget.org/packages/MinVer/1.0.0-alpha.9).

PR #60, opened 2018-10-31
suppress MinVer for non-SDK projects and print warning

### LINKED ISSUE #59: Minver doesn't work with the legacy project format
While investigating stuff for #55, I realized that MinVer currently only works when using it with an SDK-style project.

The reason is that SDK-style projects will generate all of the assembly version attributes for you based on the `Version` MSBuild property.

Legacy projects don't do that, so the version doesn't propagate to the assembly at all. If we wanted to support this, we'd have to generate a .cs file with all of the required attributes.

Otherwise, you could just document SDK-style projects as another requirement and then nothing else needs to be done. It might be a good idea to add something to the targets file that would actually cause a build warning/error if it's detected it's being used from a legacy project, instead of it just silently not doing anything.

Relates to #3.

adamralph:
Fixes #59.

Connects to #3.

### DIFF
--- a/MinVer/MinVer.csproj
+++ b/MinVer/MinVer.csproj
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <Description>Minimalistic versioning for .NET projects.</Description>
+    <Description>Minimalistic versioning for .NET SDK-style projects.</Description>
     <DevelopmentDependency>true</DevelopmentDependency>
     <IncludeBuildOutput>false</IncludeBuildOutput>
     <LangVersion>latest</LangVersion>
--- a/MinVer/MinVer.targets
+++ b/MinVer/MinVer.targets
@@ -21,12 +21,16 @@
     </PropertyGroup>
   </Target>
 
-  <Target Name="MinVer" BeforeTargets="CoreCompile;GenerateNuspec" DependsOnTargets="MinVer_GetVersion" Condition="'$(DesignTimeBuild)' != 'true'">
+  <Target Name="MinVer" BeforeTargets="CoreCompile;GenerateNuspec" DependsOnTargets="MinVer_GetVersion" Condition="'$(UsingMicrosoftNETSdk)' == 'true' And '$(DesignTimeBuild)' != 'true'">
     <PropertyGroup>
       <MinVerVersion Condition=" '$(MinVerBuildMetadata)' != '' ">$(MinVerVersion)+$(MinVerBuildMetadata)</MinVerVersion>
       <Version>$(MinVerVersion)</Version>
       <PackageVersion>$(MinVerVersion)</PackageVersion>
     </PropertyGroup>
+  </Target>
+
+  <Target Name="MinVer_Warnings" BeforeTargets="MinVer" Condition="'$(DesignTimeBuild)' != 'true'">
+    <Error Condition="'$(UsingMicrosoftNETSdk)' != 'true'" Code="MINVER0002" Text="MinVer only works in SDK-style projects." />
   </Target>
 
 </Project>
--- a/README.md
+++ b/README.md
@@ -5,11 +5,11 @@
 _[![nuget](https://img.shields.io/nuget/v/MinVer.svg?style=flat)](https://www.nuget.org/packages/MinVer)_
 _[![Build status](https://ci.appveyor.com/api/projects/status/0ai8j3x4tg6w3ima/branch/master?svg=true)](https://ci.appveyor.com/project/adamralph/min-ver/branch/master)_
 
-A minimalistic [.NET package](https://www.nuget.org/packages/MinVer) for versioning .NET projects using Git tags.
+A minimalistic [.NET package](https://www.nuget.org/packages/MinVer) for versioning .NET SDK-style projects using Git tags.
 
 ## Prerequisites
 
-- [.NET Core SDK](https://www.microsoft.com/net/download)
+- [.NET Core SDK 2.1.300 or later](https://www.microsoft.com/net/download)
 - [libcurl](https://curl.haxx.se/libcurl/) (Linux only)
 
 ## Quick start


adamralph:
@bording how does this look as a fix for #59?

MinVer/MinVer.targets
```diff
@@ -21,12 +21,16 @@
     </PropertyGroup>
   </Target>
 
-  <Target Name="MinVer" BeforeTargets="CoreCompile;GenerateNuspec" DependsOnTargets="MinVer_GetVersion" Condition="'$(DesignTimeBuild)' != 'true'">
+  <Target Name="MinVer" BeforeTargets="CoreCompile;GenerateNuspec" DependsOnTargets="MinVer_GetVersion" Condition="'$(UsingMicrosoftNETSdk)' == 'true' And '$(DesignTimeBuild)' != 'true'">
```
bording: There's no harm in having this target run, so I'm not sure the extra condition is necessary.

MinVer/MinVer.targets
```diff
@@ -21,12 +21,16 @@
...
     <PropertyGroup>
       <MinVerVersion Condition=" '$(MinVerBuildMetadata)' != '' ">$(MinVerVersion)+$(MinVerBuildMetadata)</MinVerVersion>
       <Version>$(MinVerVersion)</Version>
       <PackageVersion>$(MinVerVersion)</PackageVersion>
     </PropertyGroup>
   </Target>
 
+  <Target Name="MinVer_Warnings" BeforeTargets="MinVer" Condition="'$(DesignTimeBuild)' != 'true'">
+    <Message Condition="'$(UsingMicrosoftNETSdk)' != 'true'" Text="WARNING: MinVer only works for Microsoft.NET.Sdk projects." Importance="high" />
```
bording: Perhaps the Warning task would be better here? https://docs.microsoft.com/en-us/visualstudio/msbuild/warning-task?view=vs-2017

MinVer/MinVer.targets
adamralph: > There's no harm in having this target run, so I'm not sure the extra condition is necessary.

I did consider that, but I'm not really it's good to have "half" of MinVer working. I guess the NuGet package might still get the right version number? But anyway, it's untested, and I'm not sure I want to deal with bugs raised around usage in legacy csproj. I think it's OK to support SDK projects only and prevent anything from happening in legacy csproj.

> If DependsOnTargets are evaluated first, then someone without the .NET Core SDK installed using a legacy project could get an error from the Exec task before seeing the warning.

I don't think this problem exists. I've tested what happens in a legacy project and I can confirm that `MinVer_Warnings` is run, but neither `MinVer` nor `MinVer_GetVersion` are run. That kind of makes sense to me, since the `BeforeTargets` is saying _just_ that, i.e. run _before_ that target is run. `DependsOnTargets` is saying, _if this target is run_, run this other target first.

MinVer/MinVer.targets
adamralph: 👍 absolutely. I'll change it.

adamralph: I've amended the commit.

MinVer/MinVer.targets
adamralph: Oh wait, I see what you're saying. If the condition was taken out, would it be a problem? I guess it would be, since we're not checking whether they have the SDK installed.

adamralph: But what I said still holds, `MinVer_Warnings` runs before `MinVer_GetVersion`, so the exact problem you describe wouldn't happen.

adamralph: @bording I'm about to release alpha 9. Do you have any more thoughts on this, or do you think it's good to go? There's no rush, it can wait for alpha 10.

bording: I'd like to validate a couple of things first, so alpha 10 sounds ok to me.

bording: After looking at this, I'm now convinced that adding the extra condition is the right call since it prevents any problems with not having the .NET Core SDK installed.

README.md
```diff
@@ -5,11 +5,11 @@
 _[![nuget](https://img.shields.io/nuget/v/MinVer.svg?style=flat)](https://www.nuget.org/packages/MinVer)_
 _[![Build status](https://ci.appveyor.com/api/projects/status/0ai8j3x4tg6w3ima/branch/master?svg=true)](https://ci.appveyor.com/project/adamralph/min-ver/branch/master)_
 
-A minimalistic [.NET package](https://www.nuget.org/packages/MinVer) for versioning .NET projects using Git tags.
+A minimalistic [.NET package](https://www.nuget.org/packages/MinVer) for versioning .NET SDK projects using Git tags.
```
bording:
```suggestion
A minimalistic [.NET package](https://www.nuget.org/packages/MinVer) for versioning .NET SDK-style projects using Git tags.
```

README.md
```diff
@@ -5,11 +5,11 @@
 _[![nuget](https://img.shields.io/nuget/v/MinVer.svg?style=flat)](https://www.nuget.org/packages/MinVer)_
 _[![Build status](https://ci.appveyor.com/api/projects/status/0ai8j3x4tg6w3ima/branch/master?svg=true)](https://ci.appveyor.com/project/adamralph/min-ver/branch/master)_
 
-A minimalistic [.NET package](https://www.nuget.org/packages/MinVer) for versioning .NET projects using Git tags.
+A minimalistic [.NET package](https://www.nuget.org/packages/MinVer) for versioning .NET SDK projects using Git tags.
 
 ## Prerequisites
 
-- [.NET Core SDK](https://www.microsoft.com/net/download)
+- [.NET Core SDK  2.1 or later](https://www.microsoft.com/net/download)
```
bording:
```suggestion
- [.NET Core SDK 2.1.300 or later](https://www.microsoft.com/net/download)
```

MinVer/MinVer.csproj
```diff
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <Description>Minimalistic versioning for .NET projects.</Description>
+    <Description>Minimalistic versioning for .NET SDK projects.</Description>
```
bording:
```suggestion
    <Description>Minimalistic versioning for .NET SDK-style projects.</Description>
```

MinVer/MinVer.targets
```diff
@@ -21,12 +21,16 @@
...
+  <Target Name="MinVer" BeforeTargets="CoreCompile;GenerateNuspec" DependsOnTargets="MinVer_GetVersion" Condition="'$(UsingMicrosoftNETSdk)' == 'true' And '$(DesignTimeBuild)' != 'true'">
     <PropertyGroup>
       <MinVerVersion Condition=" '$(MinVerBuildMetadata)' != '' ">$(MinVerVersion)+$(MinVerBuildMetadata)</MinVerVersion>
       <Version>$(MinVerVersion)</Version>
       <PackageVersion>$(MinVerVersion)</PackageVersion>
     </PropertyGroup>
   </Target>
 
+  <Target Name="MinVer_Warnings" BeforeTargets="MinVer" Condition="'$(DesignTimeBuild)' != 'true'">
+    <Warning Condition="'$(UsingMicrosoftNETSdk)' != 'true'" Text="MinVer only works in Microsoft.NET.Sdk projects." />
```
bording:
```suggestion
    <Warning Condition="'$(UsingMicrosoftNETSdk)' != 'true'" Text="MinVer only works in SDK-style projects." />
```

bording: I'm also wondering if this should be an error instead of a warning.

adamralph: I think it should be. Effectively, the error will be saying "this is wrong, fix it", which I think is reasonable, since there is no reason to have MinVer installed in a legacy project. I'll change it.

adamralph:
@bording thanks for all your help with this PR! FYI I squashed all the commits into single, co-authored commit.

PR #65, opened 2018-11-02
move MinVer.Cli to tools folder

### LINKED ISSUE #64: "runtimes" folder isn't a valid TFM
I'm not sure if this causes any actual problem or not, but the current folder structure of the package might not be ideal:

image

NuGet expects folders underneath the `build` folder to be valid TFMs.

It might be worth moving all of the MinVer.Cli bits into a new `MinVer` root folder in the package, and then update the path in MinVer.targets to be aware of the change.

adamralph:
Fixes #64

### DIFF
--- a/MinVer/MinVer.csproj
+++ b/MinVer/MinVer.csproj
@@ -23,14 +23,9 @@
   </ItemGroup>
 
   <ItemGroup>
-    <None Update="MinVer.targets">
-      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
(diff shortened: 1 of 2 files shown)


bording:
Moving over to the tools folder works. I know Visual Studio/NuGet do some special things with the tools folder, like add it to the PATH of the Package Manger Console window, but I can't how that would would be a problem for this.

adamralph:
Yeah, I saw something about that too, but figured it would be OK.

PR #70, opened 2018-11-03
switch from tools/ to MinVer/ inside package

### LINKED ISSUE #64: "runtimes" folder isn't a valid TFM

adamralph:
Fixes #64


adamralph:
Released in [alpha 11(https://www.nuget.org/packages/MinVer/1.0.0-alpha.11).

PR #72, opened 2018-11-03
show build metadata in messages

adamralph:


adamralph:
@bording I think this is low risk enough to go straight into the first beta. Would you agree?

All it does is move the adding of the build metadata from the targets file to the library (which is probably where it always should have been). The functional effect is that the build metadata is now shown in the version when verbose logging is switched on.

bording:
@adamralph I agree, first beta seems fine.

adamralph:
@bording I've just realised that this introduced another behavioural change. Previously, when `MinVerVersion` was set, the targets file still appended `MinVerBuildMetadata` to it, but after this change, that no longer happens. I.e. `MinVerVersion` is now a _complete_ override, including build metadata, whereas before it was only a partial override for the non-build metadata part.

That made me ask whether the old behaviour should be preserved, but if so, it raises other questions: What if `MinVerVersion` is specified _with_ build metadata and `MinVerBuildMetadata` is also specified? An exception could be thrown, or the `MinVerVersion` build metadata could take precedence, but then how can I specify `MinVerVersion` with no build metadata but also specify _not_ to append `MinVerBuildMetadata`? This leads me to think that the new behaviour is better, but it should be made explicit in the FAQ that `MinVerVersion` must also include any desired build metadata.

What are your thoughts?

adamralph:

bording:
Seeing how this is positioned as an override to use when something is wrong or you otherwise want to disable MinVer, I think it's fine for build metadata to need to be manually specified as part of the override.

No objections to this change.

The only thing I wonder is if the env. variable should be named something like `MINVER_VERSION_OVERRIDE` to make it more explicit as to the purpose.

adamralph:
Thanks @bording.

> The only thing I wonder is if the env. variable should be named something like `MINVER_VERSION_OVERRIDE` to make it more explicit as to the purpose.

I'll give that some thought.

PR #77, opened 2018-11-03
add a message indicating that the version is overridden

adamralph:
Relates to #3.


adamralph:

PR #82, opened 2018-11-04
simplify inputs

adamralph:
Relates to #3.

### DIFF
--- a/MinVer.Cli/Program.cs
+++ b/MinVer.Cli/Program.cs
@@ -12,39 +12,39 @@ namespace MinVer.Cli
             app.HelpOption();
 
             var buildMetadata = app.Option("-b|--build-metadata <BUILD_METADATA>", "The build metadata to append to the version.", CommandOptionType.SingleValue);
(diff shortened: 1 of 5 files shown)


adamralph:
@bording after more thought, and after feedback in https://github.com/FakeItEasy/FakeItEasy/pull/1484, I decided to simplify the inputs a little, so that the user is guided more towards what should be in an environment variable and what should be in an MSBuild property.

I also decided to rename `MinVerMinimumMajorMinor` to `MinVerMajorMinor` since the former is a little unwieldy, and it suggests that something else (i.e. a tag) should be moving the version further forward, with this property purely indicating a _minimum_. Under normal usage, it _is_ the current major minor range, even if a later tag will override it. I.e. when moving to a new range, the _correct_ thing to do is to change that value rather than push a later tag. If you are going to push a later tag, it implies that you're not really using that value anyway, and you should remove it and just use tags throughout.

Interested to know your thoughts.

BTW - I've decided to do at least another round of alpha, but perhaps just on my own projects. There's been enough change since alpha 12 that I'm not confident of advertising a beta straight away.

adamralph:

bording:
This all makes sense to me. :+1:

PR #86, opened 2018-11-05
remove MINVER_TAG_PREFIX and add more guidance around env vars vs properties

adamralph:
Relates to #3.

### DIFF
--- a/MinVer/build/MinVer.targets
+++ b/MinVer/build/MinVer.targets
@@ -2,7 +2,6 @@
 
   <PropertyGroup>
     <MinVerMajorMinor Condition=" '$(MinVerMajorMinor)' == '' ">0.0</MinVerMajorMinor>
-    <MinVerTagPrefix Condition=" '$(MinVerTagPrefix)' == '' ">$(MINVER_TAG_PREFIX)</MinVerTagPrefix>
     <MinVerVerbose Condition=" '$(MinVerVerbose)' == '' ">$(MINVER_VERBOSE)</MinVerVerbose>
     <MinVerVerbose Condition=" '$(MinVerVerbose)' == '' ">false</MinVerVerbose>
     <MinVerVersion Condition=" '$(MinVerVersion)' == '' ">$(MINVER_VERSION)</MinVerVersion>
--- a/README.md
+++ b/README.md
@@ -11,6 +11,7 @@ A minimalistic [.NET package](https://www.nuget.org/packages/MinVer) for version
 - [Quick start](#quick-start)
 - [Usage](#usage)
 - [FAQ](#faq)
+- [Settings reference](#settings-reference)
 
 ## Prerequisites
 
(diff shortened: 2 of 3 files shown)


bording:
Looking at how complex it is to explain the MSBuild vs. env variable, and all the casing differences, I'm wondering if it might be better to just drop all of the special `ALL_CAPS` versions and just document that you can set any of the properties as env variables as well since that is just standard MSBuild stuff.

adamralph:
@bording I agree with you about the complexity of the explanation. Take a look at the latest version of the commit. I decided to burn it. 🔥 

The only synonyms left are `MINVER_VERBOSE` and `MinVerVerbose`. I believe the explanation isn't required for them. It's almost certain that, according to taste, you'll either set it to `true` in the project and never override it, or occasionally set it `true` on the build server when troubleshooting.

Note that one concrete reason for having both an env var name and a different property name for the same setting is to allow the env var to override the property. Typically that would done by temporarily adding an env var in a build server UI. If you don't have two different names, there's no way to recognise an override, since the property in the users project will override the env var. The user would have to code the override themselves with an appropriate MSBuild condition, but that's already going past "vanilla" XML and into actual MSBuild incantations, which is a nasty burden to lay on the user. It also requires them to have thought of that up-front.

Overriding is somewhat of a moot point for `MINVER_VERBOSE` and `MinVerVerbose` since setting the property to `false` in the project would be weird, as would overriding `true` in the project with `false` on the build server.

Right now I can't think of any use cases for the other variables being available in both variants:

- Build metadata in a property. 👎 It's specific to the build environment, not the commit.
- The major minor range in an env var. 👎 It's specific to the commit, not the build environment.
- The tag prefix in an env var. 🤔 OK, I was _slightly_ hesitant about this one, but generally the tag prefix is specific to the repo, not the build environment. I guess you might decide to _change_ your tag prefix at some point, but then in order to version properly according to history, you might have to go back and add new tags next to all the old tags. In that case, there _could_ be an argument for putting this into an env var on the build server in order to be able to build older commits, but this a sharp edge case, and a weird scenario which would probably carry other problems with it. 👎 I'm not convinced the benefit outweighs the risk of guiding the user down the wrong path by offering this as an env var.
- The version override in a property. 👎 If you're going to push a commit, you may as well replace the MinVer package reference with a `Version` property.

Of course, ~dangerous~ advanced users can figure out themselves that they can provide any of the settings in either properties or env vars right now, and they can code their own overrides, but assuming my analysis of the use cases is correct, they would only be doing that to satisfy some strange edge case.

Until there is a solid use case for overrides, your suggestion of dropping the upper case variants does hold some water. But as usual with these things, you only find out about the use cases you didn't think of once people start using the package. The problem is, if the other variant of a setting needs to be introduced later, and the current variant is not canonically used in either the project or in env vars, it's difficult to introduce the new variant in a way that makes sense. If you introduced it as an "override" it wouldn't make sense for anyone using the current variant in an env var, and likewise if you introduce an "overridable" (not to mention the naming), it wouldn't make sense for anyone using the current variant in a property. If either of those users then discovered that they _do_ need any new overrides, they would have to start shuffling their current setting from env vars into source code or vice-versa. And sooner or later, someone will put an override in a property and an overrideable in env var and wonder what's going on.

The more I think about it, the more I think that the fact these settings can be specified as either env vars or properties is actually a hindrance in the context of a targets package like MinVer. It would be simpler if MinVer could enforce what can be set in an env var and what can be set in a property, but I'm not sure that's possible. I guess, for a given env var, it could be read directly, ignoring properties, but for a given property, I'm not sure there is a way to ignore it if it came from an env var.

adamralph:
@bording I'm merging this PR since it's now become a simplification, but I'd be very interested to hear your thoughts on the env var/property topic.

bording:
> Note that one concrete reason for having both an env var name and a different property name for the same setting is to allow the env var to override the property. Typically that would done by temporarily adding an env var in a build server UI. If you don't have two different names, there's no way to recognise an override, since the property in the users project will override the env var

That's not the only way you could do that. A MinVer.props file would be imported before the content of the csproj, so you could check the property value there to see if it already has a value, which would mean it came in via env variable. You could then put that into a separate property, allowing that value to take precedence over anything specified in the csproj.

I'm not sure that would be better, but it would allow for a unification.

>T he more I think about it, the more I think that the fact these settings can be specified as either env vars or properties is actually a hindrance in the context of a targets package like MinVer.

I think you could probably do something like this by using that same MinVer.props file to ensure all the properties are empty, and haven't already been set to something from env variables.

In general, I don't really like that env variables get automatically picked up. I know there has been some discussion about being able to disable that, but I don't know if that's actually being worked on yet.

adamralph:
Released in [alpha 14](https://www.nuget.org/packages/MinVer/1.0.0-alpha.14).

PR #90, opened 2018-11-07
nicer arg errors

adamralph:
Relates to #3.

## Before

image

## After

image

Unfortunately the second error can't be avoided without using a custom task.


adamralph:
Released in [alpha 15](https://www.nuget.org/packages/MinVer/1.0.0-alpha.15).

PR #96, opened 2018-11-14
overhaul options

adamralph:
relates to #3


bording:
🤔 I'll need to take a close look at this later, but initial reaction is that this feels like too much and there's likely a simpler way.

adamralph:
Thanks @bording, as usual your opinion would be valued.

adamralph:
BTW the precedence was the wrong way round. I've fixed it in #98.

adamralph:
Superseded by #102.

PR #97, opened 2018-11-14
convert MinVer CLI to a proxy task lib

adamralph:
Relates to #3.

PR #99, opened 2018-11-14
tasks can't have empty properties

adamralph:
relates to #97


adamralph:
Released in [alpha 15](https://www.nuget.org/packages/MinVer/1.0.0-alpha.15).

PR #101, opened 2018-11-14
Miscellaneous project cleanup

bording:
This cleans up a few things I noticed while looking at the solution.


MinVer/MinVer.csproj
```diff
@@ -20,9 +20,11 @@
   </ItemGroup>
 
   <ItemGroup>
+    <None Remove="build\**\*" />
+    <None Remove="buildMultiTargeting\**\*" />
```
adamralph: What does this do?

PR #105, opened 2018-11-18
rename projects

### LINKED ISSUE #103: Rename projects and packages
From https://github.com/adamralph/min-ver/pull/100#issuecomment-438874830 by @bording:

> Right now, the MinVer.Tasks project is the actual commandline tool invoked by MSBuild, and it has a project reference to MinVer. However, MinVer is the project that has the MSBuild targets, requires the publish output from MinVer.Tasks, and builds the package, so it's actually a bit of a circular dependency. The MSBuild stuff is also split between two projects.
> 
> Instead, I think we should turn MinVer into something like MinVer.Core, and it just has the actual versioning logic in it. Then we need an additional project for each of the "shipping vehicles" we want, in this case, something running inside of MSBuild, and the commandline tool being proposed by this PR.
> 
> At the moment, the MSBuild project would also have a dependency on the commandline tool, but we should still be able to make that work. In the future, the MSBuild project should probably just become an MSBuild task, and then that dependency is removed.

adamralph:
Closes #103.


adamralph:
@bording how does this look?

MinVer/MinVer.csproj
```diff
@@ -5,26 +5,33 @@
...
     <PackageProjectUrl>https://github.com/adamralph/min-ver</PackageProjectUrl>
     <PackageReleaseNotes>https://github.com/adamralph/min-ver/milestones?state=closed</PackageReleaseNotes>
     <PublishRepositoryUrl>true</PublishRepositoryUrl>
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="LibGit2Sharp" Version="0.25.3" PrivateAssets="All" />
+    <PackageReference Include="LibGit2Sharp" Version="0.25.3" IncludeAssets="compile" />
```
bording: Something is wrong here. These dependencies need to be bundled into the package and can't be a package dependency.

bording:
While working on #106, I also noticed that the code analyzer you have in the project doesn't like the `MinVer.Lib` project name because `Lib` is a reserved keyword in some languages. Maybe drop the period and change it to `MinVerLib` instead?

adamralph:
Thanks for the review @bording, and for #106!

> Should package descriptions be updated to distinguish the packages?

👍 I'll push something.

> There are package references that can possibly be removed and instead rely on transitive references

I guess you addressed all that in #106?

> There are things that are shared by most/all projects that could be moved into Directory.Build.props

Yeah, I guess these could go in there at least:

```xml
    <PackageReference Include="Microsoft.CodeQuality.Analyzers" Version="2.6.2" PrivateAssets="All" />
    <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0-beta-63127-02" PrivateAssets="All" />
```

But I think it would need the introduction of `src` and `test` folders since I'm not sure I want those in the test and targets projects. Whatever the solution is, I think it's best done in a separate PR.

> While working on #106, I also noticed that the code analyzer you have in the project doesn't like the MinVer.Lib project name because Lib is a reserved keyword in some languages. Maybe drop the period and change it to MinVerLib instead?

Yeah, that's been bothering me. I'm not wild about `MinVerLib`, but I'm also not too bothered about either the eventual name or the warning while the assembly is an internal asset. I think I'll merge this as-is, and we can address this in another PR.

adamralph:
Released in [alpha 15](https://www.nuget.org/packages/MinVer/1.0.0-alpha.15), and as [CLI tool](https://www.nuget.org/packages/minver-cli/1.0.0-alpha.15).

bording:
> I guess you addressed all that in #106?

Yep!

> But I think it would need the introduction of src and test folders since I'm not sure I want those in the test and targets projects. Whatever the solution is, I think it's best done in a separate PR.

`src` and `test` folders sound like a good way to go.

> It's a shame about this, but I guess there's nothing that can be done (other than something terribly hacky involving tampering with the nupkg zipped contents) until NuGet/Home#5132 is resolved. 🤷‍♂️

The other thing you could is directly create the nuspec file and feed that into the project system's packing process, which let's you customize the output to be exactly what you want.

The downside of course, is that you'd then have more manual maintenance of that file, so I don't really recommend going this way.

Issue #112, opened 2018-11-20
[2025-11-10] Use a custom MSBuild task


adamralph:
This will replace the use of the `Exec` task to run `MinVer.dll` as an executable, with the use of a custom task within that assembly. This is dependent on ~either #256, or changes in `LibGit2Sharp` which @bording is aware of~ https://learn.microsoft.com/en-us/dotnet/core/whats-new/dotnet-10/sdk#use-net-msbuild-tasks-with-net-framework-msbuild, which means it is on hold until SDK until the .NET 10 SDK is the earliest supported version ([2026-11-10](https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core#lifecycle)).

One functional effect of this change is that the MSBuild verbosity level will be respected, which means that `MinVerVerbosity` will be ignored (probably deprecated initially, with a warning message). This could be considered a breaking change. It will also mean that the calculated version will not be outputted to the console (when the MSBuild verbosity level is `detailed`) as a side effect of using the `Exec` task.

adamralph:
I've been thinking about the pros and cons of introducing this in 1.0.0, but still delegating to an executable for now. In order to do proper logging, the executable can write to stderr with prefixes that indicate the log level, so that the task can parse them and propagate them back to the MSBuild logger.

## Pros

- removes stdout noise
- removes the need for `MinVerVerbosity` (the logging level will be inherited from MSBuild)
- nicer errors than with the `Exec` task
- simplifies targets file

## Cons

- involves managing System.Process
- requires listening to stderr and parsing (I'm not sure how reliable this is)
- requires another project sitting between MinVer and MinVer.Lib, and more complex packaging

bording:
> involves managing System.Process
>requires listening to stderr and parsing (I'm not sure how reliable this is)

There is a [ToolTask](https://docs.microsoft.com/en-us/dotnet/api/microsoft.build.utilities.tooltask?view=netframework-4.7.2) type that would be worth looking at to see if it helps with these.

> requires another project sitting between MinVer and MinVer.Lib, and more complex packaging

It seems like it could be possible to use the minver-cli output instead of needing a separate project.

adamralph:
One thing to bear in mind is that there is no such thing as a "current log level" in the context of an MSBuild task, since the host process can have many different loggers attached to it, each configured to a different level. That means that every message, including all debug and trace messages, will have to be output to stderr, and parsed, on every run.

[...]

PR #113, opened 2018-11-20
[WIP] output version number parts as properties

### LINKED ISSUE #109: Output the version number parts as properties
E.g. `MinVerMajor`, `MinVerMajor`, `MinVerMinor`.

These can be used for things like fixing the [assembly version](https://docs.microsoft.com/en-ca/dotnet/standard/library-guidance/versioning#assembly-version) to the `{MinVerMajor}.0.0`, or setting the [assembly file version](https://docs.microsoft.com/en-ca/dotnet/standard/library-guidance/versioning#assembly-version) to `{MinVerMajor}.{MinVerMajor}.{MinVerMinor}.{build number}`.

adamralph:
@bording this is a horrid attempt at solving #109.

Functionally, it's fine, although the downside is a whole lot more noise in the build output. The stdout side-effect of using the `Exec` task goes from one line to six (although this will go away with #112):

```
Version:2.0.0-alpha.0.1+build.42
VersionPrefix:2.0.0
VersionSuffix:alpha.0.1
Major:2
Minor:0
Patch:0
```

Code-wise, it's awful. Specially since taking a substring directly on `@(MinVerOutputVersion)`, etc. didn't work (or at least my MSBuild-fu failed me) so I had to do that in a second series of property assignments.

Can you think of a better way? 🤞

One more thing is that this may also require a re-consideration of the `MinVerVersionOverride` scenario, since currently, the only properties which would be set are `Version` and `PackageVersion`. Either the README would need to be updated to reflect the added burden on the user to set `VersionPrefix` and `VersionSuffix, `MinVerMajor`, `MinVerMinor`, and `MinVerPatch` manually, which is horrible, or the version override should be fed into `MinVer.dll` so that it can be pulled apart into those various properties. This is marked as WIP on that basis.

Closes #109.

### DIFF
--- a/MinVer.Lib/Version.cs
+++ b/MinVer.Lib/Version.cs
@@ -9,9 +9,6 @@ namespace MinVer.Lib
     public class Version : IComparable<Version>
 #pragma warning restore CA1036 // Override methods on comparable types
     {
-        private readonly int major;
-        private readonly int minor;
-        private readonly int patch;
         private readonly List<string> preReleaseIdentifiers;
         private readonly int height;
         private readonly string buildMetadata;
@@ -22,32 +19,44 @@ namespace MinVer.Lib
 
         private Version(int major, int minor, int patch, IEnumerable<string> preReleaseIdentifiers, int height, string buildMetadata)
         {
-            this.major = major;
-            this.minor = minor;
-            this.patch = patch;
+            this.Major = major;
+            this.Minor = minor;
+            this.Patch = patch;
             this.preReleaseIdentifiers = preReleaseIdentifiers?.ToList() ?? new List<string>();
             this.height = height;
             this.buildMetadata = buildMetadata;
         }
 
+        public int Major { get; }
+
+        public int Minor { get; }
+
+        public int Patch { get; }
+
+        public string ToPrefix() =>
+            $"{this.Major}.{this.Minor}.{this.Patch}";
+
+        public string ToSuffix() =>
+            $"{string.Join(".", this.preReleaseIdentifiers)}{(this.height == 0 ? "" : $".{this.height}")}";
+
         public override string ToString() =>
-            $"{this.major}.{this.minor}.{this.patch}{(this.preReleaseIdentifiers.Count == 0 ? "" : $"-{string.Join(".", this.preReleaseIdentifiers)}")}{(this.height == 0 ? "" : $".{this.height}")}{(string.IsNullOrEmpty(this.buildMetadata) ? "" : $"+{this.buildMetadata}")}";
+            $"{this.ToPrefix()}{(this.preReleaseIdentifiers.Count == 0 ? "" : $"-{this.ToSuffix()}")}{(string.IsNullOrEmpty(this.buildMetadata) ? "" : $"+{this.buildMetadata}")}";
 
         public int CompareTo(Version other)
         {
-            var major = this.major.CompareTo(other.major);
+            var major = this.Major.CompareTo(other.Major);
             if (major != 0)
             {
                 return major;
             }
 
-            var minor = this.minor.CompareTo(other.minor);
+            var minor = this.Minor.CompareTo(other.Minor);
             if (minor != 0)
             {
                 return minor;
             }
 
-            var patch = this.patch.CompareTo(other.patch);
+            var patch = this.Patch.CompareTo(other.Patch);
             if (patch != 0)
             {
                 return patch;
@@ -99,16 +108,16 @@ namespace MinVer.Lib
 
         public Version WithHeight(int height) =>
             this.preReleaseIdentifiers.Count == 0 && height > 0
-                ? new Version(this.major, this.minor, this.patch + 1, new[] { "alpha", "0" }, height, default)
-                : new Version(this.major, this.minor, this.patch, this.preReleaseIdentifiers, height, height == 0 ? this.buildMetadata : default);
+                ? new Version(this.Major, this.Minor, this.Patch + 1, new[] { "alpha", "0" }, height, default)
+                : new Version(this.Major, this.Minor, this.Patch, this.preReleaseIdentifiers, height, height == 0 ? this.buildMetadata : default);
 
         public Version AddBuildMetadata(string buildMetadata)
         {
             var separator = !string.IsNullOrEmpty(this.buildMetadata) && !string.IsNullOrEmpty(buildMetadata) ? "." : "";
-            return new Version(this.major, this.minor, this.patch, this.preReleaseIdentifiers, this.height, $"{this.buildMetadata}{separator}{buildMetadata}");
+            return new Version(this.Major, this.Minor, this.Patch, this.preReleaseIdentifiers, this.height, $"{this.buildMetadata}{separator}{buildMetadata}");
         }
 
-        public bool IsBefore(int major, int minor) => this.major < major || (this.major == major && this.minor < minor);
+        public bool IsBefore(int major, int minor) => this.Major < major || (this.Major == major && this.Minor < minor);
 
         public static Version ParseOrDefault(string text, string prefix) =>
             text == default || !text.StartsWith(prefix ?? "") ? default : ParseOrDefault(text.Substring(prefix?.Length ?? 0));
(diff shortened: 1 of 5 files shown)


adamralph:
Another idea is to stick with the current behaviour by default, and have a `MinVerExtendedProperties` option which can be set to `true` when these extra properties are required. The option would deprecated and ignored as part of #112.

bording:
See my comment on #109 on some thoughts about the need for these changes. Can we discuss those before I dig into the details of what this PR is doing and see if there's a better way to do it?

MinVer.Lib/Version.cs
```diff
@@ -22,32 +19,44 @@ public class Version : IComparable<Version>
...
             this.buildMetadata = buildMetadata;
         }
 
+        public int Major { get; }
+
+        public int Minor { get; }
+
+        public int Patch { get; }
+
+        public string ToPrefix() =>
```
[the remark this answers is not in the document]
adamralph: Done

MinVer.Lib/Version.cs
```diff
@@ -22,32 +19,44 @@ public class Version : IComparable<Version>
...
+            $"{this.Major}.{this.Minor}.{this.Patch}";
+
+        public string ToSuffix() =>
```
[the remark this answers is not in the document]
adamralph: Removed. Not required (yet).

MinVer/build/MinVer.targets
```diff
@@ -34,6 +50,8 @@
   <Target Name="MinVer" BeforeTargets="CoreCompile;GenerateNuspec" DependsOnTargets="MinVer_GetVersion" Condition="'$(UsingMicrosoftNETSdk)' == 'true' And '$(DesignTimeBuild)' != 'true'">
     <PropertyGroup>
       <Version>$(MinVerVersion)</Version>
+      <VersionPrefix>$(MinVerVersionPrefix)</VersionPrefix>
```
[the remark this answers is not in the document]
adamralph: Removed.

adamralph:
@bording I've amended the commit to reflect our recent discussions. I think we can leave out the pre-release identifiers until a use-case arises.

adamralph:
@bording I think the only remaining considerations here are:

> ...the downside is a whole lot more noise in the build output...

The only way to get around that is to wait for https://github.com/adamralph/minver/issues/112 before implementing these features. I think it's better to live with the noise for now.

And:

> One more thing is that this may also require a re-consideration of the `MinVerVersionOverride` scenario, since currently, the only properties which would be set are `Version` and `PackageVersion`. Either the README would need to be updated to reflect the added burden on the user to set `MinVerMajor`, `MinVerMinor`, `MinVerPatch`, and `AssemblyVersion` manually, which is at least a bit horrible, or the version override should be fed into `MinVer.dll` so that it can be pulled apart into those properties. This is marked as WIP on that basis.

I'm really not sure what the best course of action is. It's good that the override completely bypasses invocation of `MinVer.dll`, since then any bugs that arise can be temporarily routed around without having to make commits to branches (to temporarily remove MinVer). That could be a significant advantage when dealing with old support branches, etc. But if we want to preserve that, then the override becomes much more laborious. So it's a trade off between effectiveness of the override, and effort involved in specifying it.

But then this got me thinking about whether we really need an override. Or if we do, should it just be `MINVERDISABLE`, which completely switches it off. In that case, you'd have to temporarily specify `VERSION`, `PACKAGEVERSION`, and any other properties on your build server. But getting back to whether we really need an override, I'm not sure we do. After all, the version only really matters when the current commit is tagged. So to "fix" your versioning, all you need to do is push a tag. And since the logic in MinVer is so (relatively) simple, I don't think anyone will get into the "I don't know what's going on" situation described in the [FAQ](https://github.com/adamralph/minver#what-if-it-all-goes-wrong). But even if we keep that FAQ, we could probably change the answer to just "create a tag".

MinVer/build/MinVer.targets
```diff
@@ -33,6 +39,7 @@
 
   <Target Name="MinVer" BeforeTargets="CoreCompile;GenerateNuspec" DependsOnTargets="MinVer_GetVersion" Condition="'$(UsingMicrosoftNETSdk)' == 'true' And '$(DesignTimeBuild)' != 'true'">
     <PropertyGroup>
+      <AssemblyFileVersion>$(MinVerMajor).0.0.0</AssemblyFileVersion>
```
bording: The property we need to set here is `AssemblyVersion`, but if we set `AssemblyVersion`, we also need to set `FileVersion` so that it continues to be set the same way it has been before this PR.

```suggestion
      <AssemblyVersion>$(MinVerMajor).0.0.0</AssemblyVersion>
      <FileVersion>$(MinVerMajor).$(MinVerMinor).$(MinVerPatch).0</FileVersion>
```

README.md
```diff
@@ -129,12 +129,25 @@ You can also specify build metadata in a version tag. If the tag is on the curre
 
 ### Can I use the version calculated by MinVer for other purposes?
 
-Yes! MinVer sets both the `Version` and `PackageVersion` MSBuild properties. Use them in a target which runs after MinVer. E.g.
+Yes! MinVer sets the `Version`, `PackageVersion`, `AssemblyVersion`, `MinVerMajor`, `MinVerMinor`, and `MinVerPatch` MSBuild properties. You can use them, or override their values, in a target which runs after MinVer.
```
bording: Should this be updated to reflect my change of also needing to set `FileVersion`? I suppose maybe not since it's only being set to maintain the value it would have if `AssemblyVersion` wasn't set.

README.md
```diff
@@ -129,12 +129,25 @@ You can also specify build metadata in a version tag. If the tag is on the curre
 
 ### Can I use the version calculated by MinVer for other purposes?
 
-Yes! MinVer sets both the `Version` and `PackageVersion` MSBuild properties. Use them in a target which runs after MinVer. E.g.
+Yes! MinVer sets the `Version`, `PackageVersion`, `AssemblyVersion`, `MinVerMajor`, `MinVerMinor`, and `MinVerPatch` MSBuild properties. You can use them, or override their values, in a target which runs after MinVer.
+
+For example, MinVer sets `AssemblyVersion` to `{major}.0.0.0`, as recommended in the official [open source library guidance](https://docs.microsoft.com/en-ca/dotnet/standard/library-guidance/versioning#assembly-version). For projects which do not create NuGet packages, you may want to override this behaviour and populate [all four parts](https://docs.microsoft.com/en-us/dotnet/framework/app-domains/assembly-versioning#assembly-version-number) of `AssemblyVersion`. E.g. using Appveyor:
```
bording: You use `{major}.0.0.0` here, but `{MAJOR}.0.0.0` above.

targets/Targets.csproj
```diff
@@ -12,7 +12,7 @@
 
   <ItemGroup>
     <PackageReference Include="Bullseye" Version="2.2.0" />
-    <PackageReference Include="SimpleExec" Version="4.0.0-beta.1" />
+    <PackageReference Include="SimpleExec" Version="3.0.0" />
```
bording: Why was this bumped to 4.0.0-beta.1 before and now back to 3.0.0?

bording:
> But then this got me thinking about whether we really need an override. Or if we do, should it just be MINVERDISABLE, which completely switches it off. In that case, you'd have to temporarily specify VERSION, PACKAGEVERSION, and any other properties on your build server. But getting back to whether we really need an override, I'm not sure we do. After all, the version only really matters when the current commit is tagged. So to "fix" your versioning, all you need to do is push a tag. And since the logic in MinVer is so (relatively) simple, I don't think anyone will get into the "I don't know what's going on" situation described in the FAQ. But even if we keep that FAQ, we could probably change the answer to just "create a tag".

What are the scenarios where you're really be wanting to override the calculated version? If it's because MinVer is "broken", would you really need to just alter some build server settings to fix it? Why not commit some changes, like removing the package and manually versioning things in that case?

And if you just need it to be some different version for some reason, it seems like tagging the repo is simpler than going into your build server settings to add environment variables.

So, based on that, I agree that we could just remove the override and assume that if the package is installed, it should be trying to version things.

README.md
adamralph: 👍 I'll change it to upper case here.

README.md
adamralph: I'll include `FileVersion` too.

targets/Targets.csproj
adamralph: Weird. I don't know why this happened! I'll revert.

adamralph:
Released in [alpha 16](https://www.nuget.org/packages/MinVer/1.0.0-alpha.16).

PR #119, opened 2018-11-22
switch to license expression

adamralph:


adamralph:
Released in [alpha 17](https://www.nuget.org/packages/MinVer/1.0.0-alpha.17).

bording:
I think this change might have been premature. The package you pushed to nuget.org doesn't have any licensing information at all, so I suspect AppVeyor may not be running the 2.1.500 SDK yet.

However, when I built it locally, NuGet Package Explorer has an error about not using `licenseUrl` and `license` at the same time. When I open it up and look at the nuspec, it does appear to have both for some reason.

adamralph:
> The package you pushed to nuget.org doesn't have any licensing information at all, so I suspect AppVeyor may not be running the 2.1.500 SDK yet.

Yes, it appears that Appveyor [is still running 2.1.403](https://www.appveyor.com/docs/windows-images-software/#net-framework) so that explains the missing license info. If [Appveyor doesn't get upgraded soon](https://twitter.com/adamralph/status/1066015299047170048), then I'll revert the change.

> However, when I built it locally, NuGet Package Explorer has an error about not using licenseUrl and license at the same time. When I open it up and look at the nuspec, it does appear to have both for some reason.

It's a [bug in NPE](https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/issues/543) caused by a [bug in NuGet](https://github.com/NuGet/Home/issues/7547). A [workaround](https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/pull/542) has been merged (but not yet released).

PR #120, opened 2018-11-22
suppress build warnings

adamralph:


MinVer.Lib/MinVer.Lib.csproj
```diff
@@ -1,6 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
+    <NoWarn>$(NoWarn);CA1716</NoWarn>
```
adamralph: @damianh says this is OK. :trollface:

damianh: Approved.

Now I want to see style cop analysis.

PR #128, opened 2018-11-23
improve readme

adamralph:


README.md
```diff
@@ -29,12 +31,36 @@ Your project will be versioned according to the latest tag found in the commit h
...
+- `MinVerPatch`
+
+Those properties are used to set the following .NET SDK properties, satisfying the official [open-source library guidance for version numbers](https://docs.microsoft.com/en-ca/dotnet/standard/library-guidance/versioning#version-numbers):
+
+Property | Value
+-- | --
+`AssemblyVersion` | `{MinVerMajor}.0.0.0`
+`FileVersion` | `{MinVerMajor}.{MinVerMinor}.{MinVerPatch}.0`
+`PackageVersion` | `{MinVerVersion}`
+`Version` | `{MinVerVersion}`
```
adamralph: @bording do you know if setting `Version` is actually required? The [guidance](https://docs.microsoft.com/en-ca/dotnet/standard/library-guidance/versioning#version-numbers) doesn't mention it.

bording: `Version` gets used in a number of places in the SDK targets files to derive various other version values, so it does get used, but it doesn't directly map to one of the final outputs.

For example, if you stop setting `Version`, you would have to manually set `InformationalVersion`

adamralph: 👍 thanks. What about `PackageVersion`?

bording: If we were setting `Version` earlier and not in a target, `PackageVersion` would automatically be derived from `Version` as well.

adamralph: Ah yes, I remember that now from the initial experimentation. 👍

PR #129, opened 2018-11-23
Simplify targets by removing MinVer_GetVersion

bording:
With the version override gone, there's no longer a need to have separate targets, so this moves everything into the MinVer target.


adamralph:
Thanks @bording

adamralph:
Released in [alpha 19](https://www.nuget.org/packages/MinVer/1.0.0-alpha.19).

PR #130, opened 2018-11-23
collapse to a single target

adamralph:
@bording #129 got me thinking that we can even go a step further. I've smoke tested this in a non-SDK project and it seems fine.


adamralph:
Released in [alpha 19](https://www.nuget.org/packages/MinVer/1.0.0-alpha.19).

Issue #92, opened 2018-11-08
NU5105 warnings on SDK style project


Mpdreamz:
I'm getting the following warning on a repository with no tags and on `master`.

Having added `<MinVerMajorMinor>1.0</MinVerMajorMinor>`

```
C:\Program Files\dotnet\sdk\2.1.400\Sdks\NuGet.Build.Tasks.Pack\build\NuGet.Build.Tasks.Pack.targets(199,5): warning NU5105: The package version '1.0.0-alpha.0.4' uses SemVer 2.0.0 or components of SemVer 1.0.0 that are not supported on legacy clients. Change the package version to a SemVer 1.0.0 string. If the version contains a release label it must start with a letter. This message can be ignored if the package is not intended for older clients. [C:\Projects\elastic\
AssemblyRewriter\AssemblyRewriter\AssemblyRewriter.csproj]
```

https://docs.microsoft.com/en-us/nuget/reference/package-versioning#semantic-versioning-200

Full 2.0 SemVer support seems quite new, how do I enforce e.g a `-alpha<datestring>`

Mpdreamz:
```
    <MinVerMajorMinor>1.0</MinVerMajorMinor>
    <MinVerTagPrefix>v</MinVerTagPrefix>
    <MinVerPrerelease>alpha</MinVerPrerelease>
    <MINVER_BUILD_METADATA>$(MinVerPrerelease)$([System.DateTime]::UtcNow.ToString(yyyyMMddTHHmmss))</MINVER_BUILD_METADATA>
```

Produces `1.0.0-alpha.0.4+alpha20181108T115014` as expected but I am trying to find a way to generate

`1.0.0-alpha20181108T115014`

adamralph:
Hi, @Mpdreamz, thanks for raising this. Regarding the warning, the idea is that if you don't care about that, you should suppress the warning: https://github.com/adamralph/simple-exec/blob/ec943f1c19e09b72fe8265929bd0437e0b7e8a3d/SimpleExec/SimpleExec.csproj#L9

Regarding your desired versioning scheme, I can't immediately think of a way to achieve that in the current alpha. I'm offline for the next few days but I'll give it some thought and get back to you next week.

adamralph:
@Mpdreamz before we invest too much further in this, are you 100% sure you can't use a SemVer 2.0 version number? SemVer 2.0 has been supported for well over a year in all the major clients:

| Client                                    | Released |
|-------------------------------------------|----------|
[...]

PR #131, opened 2018-11-24
[WIP] put build number into FileVersion

adamralph:
WIP because README changes are required.

### DIFF
--- a/MinVer/build/MinVer.targets
+++ b/MinVer/build/MinVer.targets
@@ -25,8 +25,20 @@
       <MinVerMajor>$(MinVerVersion.Split(`.`)[0])</MinVerMajor>
       <MinVerMinor>$(MinVerVersion.Split(`.`)[1])</MinVerMinor>
       <MinVerPatch>$(MinVerVersion.Split(`.`)[2].Split(`-`)[0].Split(`+`)[0])</MinVerPatch>
+      <!--Appveyor-->
+      <MinVerBuildNumber Condition="$(MinVerBuildNumber) == ''">$(APPVEYOR_BUILD_NUMBER)</MinVerBuildNumber>
+      <!--Bamboo-->
+      <MinVerBuildNumber Condition="$(MinVerBuildNumber) == ''">$(BAMBOO_BUILDNUMBER)</MinVerBuildNumber>
+      <!--Azure DevOps-->
+      <MinVerBuildNumber Condition="$(MinVerBuildNumber) == ''">$(BUILD_BUILD_NUMBER)</MinVerBuildNumber>
+      <!--Jenkins, TeamCity-->
+      <MinVerBuildNumber Condition="$(MinVerBuildNumber) == ''">$(BUILD_NUMBER)</MinVerBuildNumber>
+      <!--Travis CI-->
+      <MinVerBuildNumber Condition="$(MinVerBuildNumber) == ''">$(TRAVIS_BUILD_NUMBER)</MinVerBuildNumber>
+      <!--default-->
+      <MinVerBuildNumber Condition="$(MinVerBuildNumber) == ''">0</MinVerBuildNumber>
       <AssemblyVersion>$(MinVerMajor).0.0.0</AssemblyVersion>
-      <FileVersion>$(MinVerMajor).$(MinVerMinor).$(MinVerPatch).0</FileVersion>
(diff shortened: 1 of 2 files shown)


adamralph:
@bording thoughts?

bording:
I personally don't see the value of putting the build number in the `FileVersion` when you can also just see it in the `InformationalVersion`, which gets displayed as "Product Version" on the Details tab.

As long as you don't mind the potential extra maintenance of needing to keep CI-specific information up to date, or getting requests to add others, then I don't really see any harm in doing this.

adamralph:
> ...when you can also just see it in the `InformationalVersion`, which gets displayed as "Product Version" on the Details tab.

Only if you are specifying `MinVerBuildMetadata`, and only if you are including the build number there. I guess there may be cases where that is undesirable, but having the build number in the `FileVersion` _is_ desirable. It's also recommended the [guidance](https://docs.microsoft.com/en-ca/dotnet/standard/library-guidance/versioning#assembly-file-version).

> ...the potential extra maintenance of needing to keep CI-specific information up to date...

I would document `MinVarBuildNumber` as a fully fledged option and an explanation of what is done with it, with a note that it is usually auto-detected in those build systems. I doubt those env vars will ever change, because it would break too many projects. Still, the workaround is easy: just set `MinVarBuildNumber`.

adamralph:
Alternatively, the auto-detection could be removed, leaving just the option.

bording:
>  It's also recommended the guidance.

Sure, I guess I just don't see the point of the guidance in this case.

Anyway, let's just assume this is something worth having. If we're going to be auto-detecting the build number from CI systems, then I'm not sure what the point of `MinVerBuildMetadata` is. Would it make more sense to pass in the detected build number and then always use that?

What about when a tag has a build section included in it? Which build number would you use then?

adamralph:
> I'm not sure what the point of `MinVerBuildMetadata` is

I've seen projects put all kinds of weird and wonderful things in there. Commit SHA, date and time, etc.

> What about when a tag has a build section included in it? Which build number would you use then?

Similarly, there's no way of telling what that build metadata contains.

bording:
The scenario I'm thinking of that we've automatically determined the build number, and that gets used in the `FileVersion`, but then some other value gets specified in `MinVerBuildMetadata`, which potentially means a build number there could different than what was specified in `FileVersion`.

I guess what I'm saying is that it seems like if we're detecting and setting the build number automatically in one place, should we be responsible for that value everywhere? That would mean setting the "build.<number>" portion of the build section (maybe when it's non-zero only) and then documenting that `MinVerBuildMetadata` is for additional metadata that you want other than build number.

adamralph:
> I guess what I'm saying is that it seems like if we're detecting and setting the build number automatically in one place, should we be responsible for that value everywhere? That would mean setting the "build." portion of the build section (maybe when it's non-zero only) and then documenting that `MinVerBuildMetadata` is for additional metadata that you want other than build number.

I had the exact same thought while writing this change. 🙂 

The reason I didn't do this too is because I thought it might be overreach for MinVer to be setting build metadata to `build.{build number}`. Some people might want completely different metadata, e.g. https://github.com/adamralph/minver/issues/92#issuecomment-436970423 and some might not want it at all. I guess `MinVerBuildMetadata` could be an override, but if someone doesn't want it all, we'd have to have another option to switch it off.

adamralph:
Come to think of it, setting `MinVerBuildNumber` to zero could function as the off-switch. And I do like the idea of a sensible default for `MinVerBuildMetadata`. Currently I'm having to repeat the same setting in all the repos which are using the alpha.

adamralph:
Out of interest, I had a look at 32 of the most popular NuGet packages in the gallery and not a single one is using build metadata in the package version.

Some of them have build metadata in `AssemblyInformationalVersion` but the spectrum of what is in there is wide and varied.

Some of them have a non-zero fourth number in `FileVersion`, but mostly just the numbers 1 or 2, sometimes 3, so not something that looks like a build number. Maybe 3 or 4 packages looked like they might have a build number in that fourth number.

On that basis, I'm dropping this.

@bording as you mentioned, `AssemblyInformationalVersion` maps to `Product version` in Windows Explorer and that can also show the build number if desired. It can also show pre-release identifiers and other build metadata, whereas `File version` is limited to the four numeric parts. I would recommend that anyone wanting to see the build number in Windows Explorer puts it somewhere in `AssemblyInformationalVersion`. Still, if anyone really wants the build number in `File version` they can write a small target to do it, as documented in the README.

PR #135, opened 2018-11-25
fix and improve logging

adamralph:
I found a bug in the trace logging in alpha 19. 😢 

It also prompted me to improve the logging.


adamralph:
Released in [alpha 20](https://www.nuget.org/packages/MinVer/1.0.0-alpha.20).

PR #141, opened 2018-11-26
Revert "refactor discovery"

adamralph:
This reverts commit 4d5da7f0d466a86a42f7673357a3d4ba9b346890.

As demonstrated in https://github.com/SQLStreamStore/SQLStreamStore.HAL/pull/38#issuecomment-441652214 there's a bug in `Repository.Discover()` which only seems to manifest in the context of Docker.

### DIFF
--- a/MinVer.Lib/RepositoryEx.cs
+++ b/MinVer.Lib/RepositoryEx.cs
@@ -1,5 +1,6 @@
 namespace MinVer.Lib
 {
+    using System.IO;
     using LibGit2Sharp;
 
     public static class RepositoryEx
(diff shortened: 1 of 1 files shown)


adamralph:
Released in [alpha 21](https://www.nuget.org/packages/MinVer/1.0.0-alpha.21).

bording:
>  there's a bug in Repository.Discover() which only seems to manifest in the context of Docker.

While reverting the change around this makes sense to unblock things, I still don't see how docker could be relevant here. 

I feel like more investigation / upstream bug filing is needed.

adamralph:
@bording I agree. Two problems for me though: I'm a Docker noob, and I have zero clue about the libgit2 native code. 🙂

FTR, it's not an OS thing, because I built other projects using alpha 20 on all OS's and they worked just fine, e.g.:

- Windows
  - 
- Linux
  - https://travis-ci.org/adamralph/bullseye/jobs/459474940#L514
  - https://travis-ci.org/adamralph/bullseye/jobs/459474940#L768
- macOS
  - https://travis-ci.org/adamralph/bullseye/jobs/459474941#L269
  - https://travis-ci.org/adamralph/bullseye/jobs/459474941#L523
- Windows
  - https://ci.appveyor.com/project/adamralph/bullseye/builds/20539822/job/jp1pir2vj7j4951a#L18
  - https://ci.appveyor.com/project/adamralph/bullseye/builds/20539822/job/jp1pir2vj7j4951a#L60

It really does seem to manifest only in Docker.

Issue #146, opened 2018-12-04
MinVerBuildMetadata not used in nupkg version


blairconrad:
Hi. I specified the `MinVerBuildMetadata` environment variable so I could [include build metadata in the version](https://github.com/adamralph/minver#can-i-include-build-metadata-in-the-version) and expected to see it appear in the nupkg version, but it does not:

```PowerShell
λ $env:MinVerBuildMetadata='fix-nupkg-output'; .\build.cmd -s pack

Bullseye: Starting... (pack) (skip dependencies)
Bullseye/pack: Starting...
Bullseye/pack/src/FakeItEasy/FakeItEasy.csproj: Starting...
dotnet pack src/FakeItEasy/FakeItEasy.csproj --configuration Release --no-build --output D:\Sandbox\FakeItEasy\artifacts\output
Microsoft (R) Build Engine version 15.8.166+gd4e8d81a88 for .NET Core
Copyright (C) Microsoft Corporation. All rights reserved.

  MinVer: MinVer 1.0.0-beta.1+build.141.fbbb85364bf594fe4a4af4f533df30acd887ec42.
  MinVer: Starting at commit 923ac0f (height 0)...
  MinVer: History diverges from 923ac0f (height 0) to:
  MinVer: - 67cfc92 (height 1)
  MinVer: - f05e091 (height 1)
  MinVer: Following path from 923ac0f (height 0) through first parent 67cfc92 (height 1)...
  MinVer: History diverges from 67cfc92 (height 1) to:
  MinVer: - 188b0af (height 2)
  MinVer: - a8cb201 (height 2)
  MinVer: Following path from 67cfc92 (height 1) through first parent 188b0af (height 2)...
  MinVer: History diverges from 188b0af (height 2) to:
  MinVer: - 31d6be6 (height 3)
  MinVer: - 033e326 (height 3)
  MinVer: Following path from 188b0af (height 2) through first parent 31d6be6 (height 3)...
  MinVer: History diverges from 31d6be6 (height 3) to:
  MinVer: - 34c291f (height 4)
  MinVer: - b6e910b (height 4)
  MinVer: Following path from 31d6be6 (height 3) through first parent 34c291f (height 4)...
  MinVer: History diverges from 34c291f (height 4) to:
  MinVer: - c83026a (height 5)
  MinVer: - 5288747 (height 5)
  MinVer: Following path from 34c291f (height 4) through first parent c83026a (height 5)...
  MinVer: History diverges from c83026a (height 5) to:
[...]

PR #160, opened 2018-12-09
log inputs and outputs

### LINKED ISSUE #147: Output all properties at detailed level
_As suggested by @blairconrad: https://github.com/adamralph/minver/issues/146#issuecomment-444124896._

adamralph:
Closes #147

### DIFF
--- a/MinVer/build/MinVer.targets
+++ b/MinVer/build/MinVer.targets
@@ -5,9 +5,19 @@
     <GetPackageVersionDependsOn>$(GetPackageVersionDependsOn);MinVer</GetPackageVersionDependsOn>
   </PropertyGroup>
 
+  <PropertyGroup>
+    <MinVerDetailed>low</MinVerDetailed>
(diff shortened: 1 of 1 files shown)


adamralph:
@blairconrad how does this look WRT https://github.com/adamralph/minver/issues/147 ?

You mentioned in https://github.com/adamralph/minver/issues/146#issuecomment-444124896 that it should also also log the nupkg filename, what you referred to as `PackageFilenameVersion`, but that is out MinVer's control. All it can do is log the properties that it sets. The derivation of the nupkg filename is done by other targets in the .NET SDK and is standard behaviour. Although there could even be other custom targets injected into the pipeline which alter it, and MinVer has no way to know about those.

adamralph:
@bording do you have any view on this?

blairconrad:
It looks good, @adamralph. I understand that the filename is out of your control (I did not then). Thanks!

adamralph:

PR #172, opened 2018-12-17
Redundant min major minor

### LINKED ISSUE #151: debug/detailed level message when MinVerMajorMinor is redundant
`MinVerMajorMinor` is redundant when a tag is found with a higher version.

Kudos to @cameronfletcher for suggesting this.

adamralph:
Closes #151

### DIFF
--- a/MinVer.Lib/Versioner.cs
+++ b/MinVer.Lib/Versioner.cs
@@ -12,6 +12,13 @@ namespace MinVer.Lib
             {
                 log.Info($"Bumping version to {calculatedVersion} to satisfy minimum major minor {minMajorMinor}.");
             }
+            else
+            {
+                if (minMajorMinor != default)
+                {
+                    log.Debug($"Minimum major minor {minMajorMinor} is redundant. The calculated version is already equal or higher.");
+                }
+            }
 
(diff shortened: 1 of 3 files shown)


bording:
Why is this needed? Given that the property is a "minumum" why do you need to warn?

Ah, but It is a debug level entry, so I guess most people won't see it.

adamralph:
@bording it's effectively the message you described in https://github.com/adamralph/minver/issues/151#issuecomment-445371035

bording:
The distinction I see between what I mentioned before and this is that by saying it's redundant, that implies it should be removed. However, given that the the property is a minimum, there's nothing wrong with having it be there. It isn't redundant, it's expected that the calculation might choose something higher than the minimum that's been specified.

adamralph:
Under normal usage, once a higher version is chosen, the property will never have any effect again since the tag used to choose the higher version will always be found in the history.

I guess where that falls down is if tags are removed, or an older tag is created in front of a newer tag. To me that is an invalid workflow, but it's difficult to second guess the exotic usage that might come to exist. To be sure, I'll take the mention of redundancy out.

adamralph:

PR #174, opened 2018-12-17
change message about min major minor from 'redundant' to 'ignoring'

adamralph:


adamralph:
@bording how does this look?

adamralph:
One thing to bear in mind is that the minimum _version_ is never specified. Only a minimum major minor range. I guess the minimum version would be something like `major.minor.0-0` but it would be quite odd to see that in the logs. What if the message said something like this?

> The calculated version {version} satisfies the minimum major minor {minMajorMinor}.

bording:
@adamralph That seems fine as well. It's just that this whole scenario doesn't fit my mental model of how I think about that property at all, which is why I'm having a disconnect here.

adamralph:
@bording OK, I've changed it to that last proposal for now. If you have some more thoughts on this, please let me know.

adamralph:
Oops, forgot to change the commit message. 🤷‍♂️

adamralph:

PR #175, opened 2018-12-18
change stdout importance to low

### LINKED ISSUE #170: Hide version from console
Currently, when building a project with MinVer installed, the version appears in output like so:

image

This is a side effect of [using the MSBuild `Exec` task](https://github.com/adamralph/minver/blob/cf677887e146fa3a0dbf822e887d135ef791ed54/MinVer/build/MinVer.targets#L28) to invoke an executable to do the versioning work.

If we manage to pull off https://github.com/adamralph/minver/issues/112, this will go away, but in the meantime, if a property could be added to the `Exec` task to suppress stdout, that would remove this side effect.

adamralph:
Closes #170.


adamralph:

Issue #171, opened 2018-12-14
Fails to build on Ubuntu


shaynevanasperen:
When building my project on Ubuntu, I get the following error:

```
Unhandled Exception: System.TypeInitializationException: The type initializer for 'LibGit2Sharp.Core.NativeMethods' threw an exception. ---> System.DllNotFoundException: Unable to load shared library 'git2-8e0b172' or one of its dependencies. In order to helpdiagnose loading problems, consider setting the LD_DEBUG environment variable: libgit2-8e0b172: cannot open shared object file: No such file or directory
     at LibGit2Sharp.Core.NativeMethods.git_libgit2_init()
     at LibGit2Sharp.Core.NativeMethods.LoadNativeLibrary()
     at LibGit2Sharp.Core.NativeMethods..cctor()
     --- End of inner exception stack trace ---
     at LibGit2Sharp.Core.NativeMethods.git_repository_open(git_repository*& repository, FilePath path)
     at LibGit2Sharp.Core.Proxy.git_repository_open(String path)
     at LibGit2Sharp.Repository..ctor(String path, RepositoryOptions options, RepositoryRequiredParameter requiredParameter)
     at MinVer.Lib.RepositoryEx.TryCreateRepo(String path, Repository& repository) in C:\projects\min-ver\MinVer.Lib\RepositoryEx.cs:line 16
     at MinVer.Program.GetVersion(String path, String tagPrefix, MajorMinor range, String buildMetadata, Verbosity level) in C:\projects\min-ver\minver-cli\Program.cs:line 80
     at MinVer.Program.<>c__DisplayClass1_0.<Main>b__0() in C:\projects\min-ver\minver-cli\Program.cs:line 34
     at MinVer.Program.Main(String[] args) in C:\projects\min-ver\minver-cli\Program.cs:line 27
  Aborted (core dumped)
```

adamralph:
@shaynevanasperen thanks for raising this. Please see the [FAQ](https://github.com/adamralph/minver#why-does-minver-fail-with-systemtypeinitializationexception).

slang25:
Hey @shaynevanasperen, you could also try referencing `LibGit2Sharp` version `0.26.0-preview-0070` as a private dependency, they have done some reduce the native dependencies in that version. I can't guarantee it will work, I'm more curious that anything.

slang25:
Try now with beta2 @shaynevanasperen

PR #177, opened 2018-12-19
Problem using minver on ubuntu 18.04

yreynhout:
- Upgraded dependencies which seems to fix the LibGit2Sharp native dll loading issue.

### DIFF
--- a/MinVer.Lib/MinVer.Lib.csproj
+++ b/MinVer.Lib/MinVer.Lib.csproj
@@ -6,9 +6,9 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="LibGit2Sharp" Version="0.25.3" />
+    <PackageReference Include="LibGit2Sharp" Version="0.26.0-preview-0070" />
     <PackageReference Include="Microsoft.CodeQuality.Analyzers" Version="2.6.2" PrivateAssets="All" />
-    <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0-beta-63127-02" PrivateAssets="All" />
+    <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0-beta2-18618-05" PrivateAssets="All" />
   </ItemGroup>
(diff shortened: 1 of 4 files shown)


targets/Targets.csproj
```diff
@@ -12,7 +12,8 @@
 
   <ItemGroup>
     <PackageReference Include="Bullseye" Version="2.3.0-beta.6" />
-    <PackageReference Include="LibGit2Sharp" Version="0.25.3" />
+    <PackageReference Include="LibGit2Sharp.NativeBinaries" Version="1.0.258" />
```
adamralph: Why is this package required?

MinVer.Lib/MinVer.Lib.csproj
```diff
@@ -6,9 +6,9 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="LibGit2Sharp" Version="0.25.3" />
+    <PackageReference Include="LibGit2Sharp" Version="0.26.0-preview-0070" />
```
adamralph: @bording what's your view on MinVer taking this pre-release dep?

targets/Targets.csproj
yreynhout: ah, it's not - I'll remove it

yreynhout: committed and squashed it

MinVer.Lib/MinVer.Lib.csproj
bording: I think it should be fine. Preview 70 is also the one that removes the curl dependency, so that is another reason to do it.

slang25: It would fix this issue: https://github.com/adamralph/minver/issues/171

adamralph: Thanks @bording.

Thanks @slang25, that's true. I've raised #179 to update the README.

adamralph:

PR #186, opened 2018-12-30
simplify build

adamralph:


slang25:
You could chose to omit `IsTestProject` from your projects, as referencing `Microsoft.NET.Test.Sdk` will set it to true for you, it will default to false otherwise.

adamralph:
@slang25 adding it explicitly to non-test projects gets rid of this message:

> Skipping running test for project C:\code\adamralph\minver\targets\Targets.csproj. To run tests with dotnet test add "<IsTestProject>true<IsTestProject>" property to project file.

Or at least, it will when this bug fix is released: https://github.com/Microsoft/vstest/issues/1866 😉

adamralph:
> referencing `Microsoft.NET.Test.Sdk` will set it to true for you

@slang25 I did not know that though. That's good to know, thanks. I'll remove it from the test projects.

Something seems off here. We shouldn't have to add it to non-test projects to get rid of that silly message.

slang25:
I hadn't realised you got a message in the log either.

I feel the message should be off by default, or at least be possible to switch off. It happens here:
https://github.com/Microsoft/vstest/blob/37696768786ef1018f0b4499eab9ca27af141057/src/Microsoft.TestPlatform.Build/Microsoft.TestPlatform.targets#L55

adamralph:
@slang25 see https://github.com/Microsoft/vstest/pull/1867/files#r240614971

slang25:
Nice, I just arrived at the same conclusion myself, you are quicker than me 😄

PR #188, opened 2019-01-01
suppress empty dependency group

adamralph:
See https://github.com/NuGet/Home/issues/5132#issuecomment-450706564

Thanks @gitfool!


adamralph:
Released in [beta 3](https://www.nuget.org/packages/MinVer/1.0.0-beta.3).

PR #210, opened 2019-02-27
Suppress NU5105

### LINKED ISSUE #209: Suppress NU5105 in consuming projects
E.g.

> C:\Program Files\dotnet\sdk\2.2.104\Sdks\NuGet.Build.Tasks.Pack\build\NuGet.Build.Tasks.Pack.targets(202,5): warning NU5105: The package version '2.0.5-alpha.0.9+build.0' uses SemVer 2.0.0 or components of SemVer 1.0.0 that are not supported on legacy clients. Change the package version to a SemVer 1.0.0 string. If the version contains a release label it must start with a letter. This message can be ignored if the package is not intended for older clients.

slang25:
Fixes #209 

Doing the easy work following @bording instruction

### DIFF
--- a/MinVer/build/MinVer.targets
+++ b/MinVer/build/MinVer.targets
@@ -8,6 +8,7 @@
   <PropertyGroup>
     <MinVerDetailed>low</MinVerDetailed>
     <MinVerDetailed Condition="'$(MinVerVerbosity)' == 'detailed' Or '$(MinVerVerbosity)' == 'd' Or '$(MinVerVerbosity)' == 'diagnostic' Or '$(MinVerVerbosity)' == 'diag'">high</MinVerDetailed>
+    <NoWarn>$(NoWarn);NU5015</NoWarn>
   </PropertyGroup>
 
   <Target Name="MinVer" BeforeTargets="CoreCompile;GenerateNuspec" Condition="'$(DesignTimeBuild)' != 'true'">


MinVer/build/MinVer.targets
```diff
@@ -46,6 +46,7 @@
       <FileVersion>$(MinVerMajor).$(MinVerMinor).$(MinVerPatch).0</FileVersion>
       <PackageVersion>$(MinVerVersion)</PackageVersion>
       <Version>$(MinVerVersion)</Version>
+      <NoWarn>$(NoWarn);NU5015</NoWarn>
```
bording: This `PropertyGroup` is inside the `Target`, which isn't what we want. Instead, if you move it up to the one on lines 8-11, and that should be perfect.

adamralph: @bording I think I also prefer it in the 8-11 block but, out of interest, would it not work inside the target?

bording: I'd have to test it to be sure, but I suspect it wouldn't because the properties in the target would only be evaluated once the target was executed, and that could be too late to influence the warning settings. 

Then even if it did work when running a build from the command line, depending on what runs during a design-time build, it might not work from inside VS.

adamralph:
Thanks @slang25 and @bording!

PR #216, opened 2019-04-04
added implementation for selective version segment incrementation

### LINKED ISSUE #212: Option to increment minor or major version instead of patch
When the latest tag is an RTM, and is on an older commit, the patch version is incremented. E.g. if the latest tag is `1.0.0` at a height of 3 commits, the current commit version is `1.0.1-alpha.0.3`.

Some projects do not use the patch version, so it would make more sense for the version to be `1.1.0-alpha.0.3`. Some projects may even not use the minor version, which means the version should be `2.0.0-alpha.0.3`.

This could be configurable via a new option.

enzian:
The when applying minver, a user should be able to override the default
behavior of how versions are incremented. The current behavior is, that
whenever minver is run on a commit, that does not carry a minver tag and
is at a height > 0 from the last minver tag, the patch version will be
incremented and the height added. For example: When the latest tag is
1.2.3 and the commit is at a height of 1 from this tag, the resolved
version will be 1.2.4-alpha.0.1.

This behavior should be customizable by setting a property that defines
what segments of the version (major, minor, patch) should be
incremented.

Closes #212

### DIFF
--- a/MinVer.Lib/Version.cs
+++ b/MinVer.Lib/Version.cs
@@ -108,10 +108,25 @@ namespace MinVer.Lib
                 ? this
                 : new Version(minMajorMinor.Major, minMajorMinor.Minor, default, new[] { "alpha", "0" }, this.height, this.buildMetadata);
 
-        public Version WithHeight(int height) =>
-            this.preReleaseIdentifiers.Count == 0 && height > 0
-                ? new Version(this.Major, this.Minor, this.Patch + 1, new[] { "alpha", "0" }, height, default)
-                : new Version(this.Major, this.Minor, this.Patch, this.preReleaseIdentifiers, height, height == 0 ? this.buildMetadata : default);
+        public Version WithHeight(int height, VersionPart autoIncrement)
+        {
+            if (this.preReleaseIdentifiers.Count == 0 && height > 0)
+            {
+                switch (autoIncrement)
+                {
+                    case VersionPart.Major:
+                        return new Version(this.Major + 1, 0, 0, new[] { "alpha", "0" }, height, default);
+                    case VersionPart.Minor:
+                        return new Version(this.Major, this.Minor + 1, 0, new[] { "alpha", "0" }, height, default);
+                    case VersionPart.Patch:
+                        return new Version(this.Major, this.Minor, this.Patch + 1, new[] { "alpha", "0" }, height, default);
+                    default:
+                        throw new ArgumentOutOfRangeException(nameof(autoIncrement));
+                }
+            }
+
+            return new Version(this.Major, this.Minor, this.Patch, this.preReleaseIdentifiers, height, height == 0 ? this.buildMetadata : default);
+        }
 
         public Version AddBuildMetadata(string buildMetadata)
         {
--- a/MinVer/Logger.cs
+++ b/MinVer/Logger.cs
@@ -48,6 +48,9 @@ namespace MinVer
         public static void ErrorRepoOrWorkDirDoesNotExist(string repoOrWorkDir) =>
             Error(1002, $"Repository or working directory '{repoOrWorkDir}' does not exist.");
 
+        public static void ErrorInvalidAutoIncrement(string autoIncrement) =>
+            Error(1006, $"Invalid auto increment '{autoIncrement}'. Valid values are {VersionPart.Patch.GetValidValues()}");
+
         public static void ErrorInvalidMinMajorMinor(string minMajorMinor) =>
             Error(1003, $"Invalid minimum MAJOR.MINOR '{minMajorMinor}'. Valid values are {MajorMinor.ValidValues}");
 
--- a/README.md
+++ b/README.md
@@ -42,7 +42,9 @@ _NOTE: The MinVer package reference should normally include `PrivateAssets="All"
 
 When the current commit is tagged with a version, the tag is used as-is.
 
-When the current commit is not tagged, MinVer searches the commit history for the latest tag. If the latest tag found is a [pre-release](https://semver.org/spec/v2.0.0.html#spec-item-9), MinVer will use it as-is. If the latest tag found is RTM (not pre-release), MinVer will increase the patch number and add default pre-release identifiers, e.g. `1.0.0` becomes `1.0.1-alpha.0`. If no tag is found, the default version `0.0.0-alpha.0` is used.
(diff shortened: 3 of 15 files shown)


enzian:
not done by a long shot, but this should serve as a basis for discussion! Also there are some new cases that we need to discuss in the `BuildMetadata.PreviousTag` test.

enzian:
There is one case I'm not sure about what to do: If we have the closes tag at `1.2.3-beta` and we increment the version, the current implementation says *not to increment at all*:
```csharp
public Version WithHeight(int height) =>
            this.preReleaseIdentifiers.Count == 0 && height > 0
                ? new Version(this.Major, this.Minor, this.Patch, new[] { "alpha", "0" }, height, default)
                : new Version(this.Major, this.Minor, this.Patch, this.preReleaseIdentifiers, height, height == 0 ? this.buildMetadata : default);
```
but this is not what I would expect the code to! I'd expect to get `1.2.4-alpha.0.1`. Is there a particular reasoning behind the current implementation?

adamralph:
@enzian if the last tag is not an RTM, e.g. `1.2.3-beta`, then the calculated version will be `1.2.3-beta.{height}`. Auto-incrementing is only done when the last tag was RTM.

This is because if I release a `1.2.3` pre-release, then the next release is still going to be in the `1.2.3` range. It's only after I release `1.2.3` RTM that I want to work on the next version, whether that's `1.2.4`, `1.3.0`, or `2.0.0`.

adamralph:
@enzian I rebased the branch and pushed a commit which simplifies the implementation a bit. Can please take a look and let me know what you think?

MinVer.Lib/Version.cs
```diff
@@ -108,9 +108,9 @@ public int CompareTo(Version other)
                 ? this
                 : new Version(minMajorMinor.Major, minMajorMinor.Minor, default, new[] { "alpha", "0" }, this.height, this.buildMetadata);
 
-        public Version WithHeight(int height) =>
+        public Version WithHeight(int height, VersionPart autoIncrement) =>
             this.preReleaseIdentifiers.Count == 0 && height > 0
-                ? new Version(this.Major, this.Minor, this.Patch + 1, new[] { "alpha", "0" }, height, default)
+                ? new Version(this.Major + (autoIncrement == VersionPart.Major ? 1 : 0), this.Minor + (autoIncrement == VersionPart.Minor ? 1 : 0), this.Patch + (autoIncrement == VersionPart.Patch ? 1 : 0), new[] { "alpha", "0" }, height, default)
                 : new Version(this.Major, this.Minor, this.Patch, this.preReleaseIdentifiers, height, height == 0 ? this.buildMetadata : default);
```
enzian: I don't find this particularly readable because of the encased conditional assignments...

README.md
```diff
@@ -222,6 +224,17 @@ The tag with the higher version is used.
 
 MinVer will use the height on the first path followed where the history diverges. The paths are followed in the same order that the parents of the commit are stored in git. The first parent is the commit on the branch that was the current branch when the merge was performed. The remaining parents are stored in the order that their branches were specified in the merge command.
 
+### Can I increment other parts of the version besides `patch`
+
+Yes! MinVer offers both a cli (`-i | --increment-segment <SEGMENT>`) option and an MSBuild property (`MinVerAutoIncrement`) to override the default behavior. The default behavior in MinVer defines, that for non-tagged commits, the `alpha.0` [suffix and the height will be appended to the version](#how-it-works). But it is possible to increment other segments of the version like the major and minor. The default for these parameters are set to `patch` and represent the default behavior. To increment the minor the value can be set to `minor` and for the major to `major`.
+
+```
+latest version    MinVerAutoIncrement   at height     resolves to
+1.2.3+a           major                 1             2.0.0-alpha.0.1
```
enzian: These cases are not in the test in `BuildMetadata.PreviousTag` anymore! Just the default case (patch) is represented there!

MinVer.Lib/Version.cs
adamralph: Yes, that's true. I always tend to over-inline. This should be split apart a bit.

README.md
adamralph: That's true, because the `BuildMetadata` class is for testing the build metadata feature. We should add a new class for testing the auto incrementing feature, which isolates test cases each of the `VersionPart` values.

MinVer.Lib/Version.cs
enzian: also, when auto incrementing, this does not reset the other (lower!?! what's the working here? :p ) segments to 0.

adamralph: Ah, good catch! 🤦‍♂️

enzian: xunit told me once I wrote the tests :smile:

adamralph:
@enzian I added some commits to tweak the tests, add a package test, and simplify the README a bit. Can you please take a look?

MinVer.Lib/Version.cs
```diff
@@ -108,10 +108,25 @@ public int CompareTo(Version other)
...
+                switch (autoIncrement)
+                {
+                    case VersionPart.Major:
+                        return new Version(this.Major + 1, 0, 0, new[] { "alpha", "0" }, height, default);
+                    case VersionPart.Minor:
+                        return new Version(this.Major, this.Minor + 1, 0, new[] { "alpha", "0" }, height, default);
+                    case VersionPart.Patch:
+                        return new Version(this.Major, this.Minor, this.Patch + 1, new[] { "alpha", "0" }, height, default);
+                    default:
+                        throw new ArgumentException();
```
adamralph: Perhaps a moot point, since this is all internal, but I guess this should really be:

```C#
throw new ArgumentOutOfRangeException(nameof(autoIncrement));
```

MinVer/Logger.cs
```diff
@@ -48,6 +48,9 @@ public void Warn(int code, string message)
         public static void ErrorRepoOrWorkDirDoesNotExist(string repoOrWorkDir) =>
             Error(1002, $"Repository or working directory '{repoOrWorkDir}' does not exist.");
 
+        public static void ErrorInvalidAutoIncrement(string autoIncrement) =>
+            Error(1006, $"Invalid auto increment '{autoIncrement}'. Valid values are {VersionPart.Patch.GetValidValues()}");
```
adamralph: This is a bit of a hack, since it's actually taking one of the enum values in order to call an extension method on it, but the alternative is something like

```C#
VersionPartExtensions.GetValidValues()
```

Given enums are value types, I think this hack is OK.

enzian: I agree, not pretty but the functionality is actually where it belongs...

MinVer.Lib/Version.cs
enzian: I agree, wait for my patch!

adamralph:
The diff looks great now @enzian. Thanks!

I'm going to squash the commits and do a final force push before merging.

enzian:
> I'm going to squash the commits and do a final force push before merging.

Sure, all good by me!

adamralph:
Thanks for all your work on this @enzian!

enzian:
@adamralph Same!

PR #218, opened 2019-04-05
Clear ItemGroups before using them

### LINKED ISSUE #217: MinVer can fail if internal item groups are already populated
This is a corner case bug that will likely never be hit, but it's simple to fix.

The internal item groups used in the MinVer target, `MinVerInputs`, `MinVerConsoleOutput`, and `MinVerOutputVersion`, are populated using the `Include` attribute. This _adds_ the specified values to the item groups. If those item groups already contain values before the `MinVer` target is executed, then the `MinVer` target can fail or produced unexpected results.

I hit this when I was putting together https://github.com/IdentityServer/IdentityServer4/pull/3163 because I was using `minver-cli` in a target which ran before the `MinVer` target and I was using the same item group names to extract the console output. It's easily worked around by using different item group names, but it caused me a huge headache trying to find out what the problem was.

bording:
Fixes #217 

I was originally going to put the `Remove` next to the first `Include` for each group, but that doesn't work for `MinVerConsoleOutput`, so I decided to add a dedicated group just for the `Remove`s instead.


adamralph:
👌 thanks!

PR #220, opened 2019-04-12
Add version information class

bording:
This adds a target that will create a static class that contains the MinVer version information, enabling programs to reference version information at runtime.

This is something that other versioning tools offer, so it seemed like something MinVer should have too!

I made it a separate target because it needs to run during design-time builds as well so intellisense works properly.

I also added a condition, so someone who doesn't want this can turn it off.

### DIFF
--- a/MinVer/build/MinVer.targets
+++ b/MinVer/build/MinVer.targets
@@ -65,4 +65,33 @@
     <Message Importance="$(MinVerDetailed)" Text="MinVer: [output] Version=$(Version)" />
   </Target>
 
+  <Target Name="MinVer_WriteVersionInformationFile" AfterTargets="MinVer" Condition="'$(MinVerCreateVersionInformationClass)' != 'false'">
+    <PropertyGroup>
+      <MinVerVersionInformationFilePath>$(IntermediateOutputPath)MinVerVersionInformation.g.cs</MinVerVersionInformationFilePath>
+      <MinVerVersionInformationFileSourceLines>
+        namespace $(RootNamespace)
+        {
+        internal static class MinVerVersionInformation
+        {
+        public const string MinVerVersion = "$(MinVerVersion)"%3b
+        public const string MinVerMajor = "$(MinVerMajor)"%3b
+        public const string MinVerMinor = "$(MinVerMinor)"%3b
+        public const string MinVerPatch = "$(MinVerPatch)"%3b
+        public const string MinVerPreRelease = "$(MinVerPreRelease)"%3b
+        public const string MinVerBuildMetadata = "$(MinVerBuildMetadata)"%3b
+        public const string AssemblyVersion = "$(AssemblyVersion)"%3b
+        public const string FileVersion = "$(FileVersion)"%3b
+        public const string PackageVersion = "$(PackageVersion)"%3b
+        public const string Version = "$(Version)"%3b
+        }
+        }
+      </MinVerVersionInformationFileSourceLines>
+    </PropertyGroup>
+    <WriteLinesToFile File="$(MinVerVersionInformationFilePath)" Lines="$(MinVerVersionInformationFileSourceLines)" Overwrite="true" WriteOnlyWhenDifferent="true" />
+    <ItemGroup>
+      <Compile Include="$(MinVerVersionInformationFilePath)" />
+      <FileWrites Include="$(MinVerVersionInformationFilePath)" />
+    </ItemGroup>
+  </Target>
+
 </Project>


bording:
If we care, we should also be able to make this work for other languages like VB and F# as well.

adamralph:
Thanks @bording. Can you describe any use cases for this? Also, what will currently happen if my project is not C#?

bording:
The use cases I can think of are around being able to use the version information in some way in your application. Things like displaying version information on a web page, or an about box, or perhaps you include version information in a message header.

For example, we would need this if we ever wanted to move NServiceBus to MinVer.

> Also, what will currently happen if my project is not C#?

At the moment I suspect it wouldn't be happy, but that is why I included the option to turn it off.  We could also condition it based on the project language, so that someone not using C# wouldn't need to manually disable it.

And we could extend that further and write something for whatever other languages that would be worth supporting.

bording:
Take a look at https://github.com/Particular/NServiceBus/pull/5360 to see how it's being used in NServiceBus.

adamralph:
@bording thanks! I'm convinced by the use cases.

> We could also condition it based on the project language, so that someone not using C# wouldn't need to manually disable it.

I think we should do that before adding this. I know there are already some F# projects using MinVer and I wouldn't want to add friction for them.

adamralph:
@bording it just occured to me that we can already use the standard reflection APIs to get all this information. E.g.

```C#
Console.WriteLine($"InformationalVersion: {this.GetType().Assembly.GetCustomAttribute<AssemblyInformationalVersionAttribute>().InformationalVersion}");
Console.WriteLine($"Version: {this.GetType().Assembly.GetName().Version}");
Console.WriteLine($"FileVersion: {this.GetType().Assembly.GetCustomAttribute<AssemblyFileVersionAttribute>().Version}");
```

The API is clunky compared to the proposed `MinVerVersionInformation` but the information is all there, it's language agnostic, and it requires no changes to MinVer.

For a nicer API, I'd rather have a dedicated runtime package that wraps around those reflection APIs and understands SemVer so that it can do things like deconstruct `InformationalVersion` into it's constituent parts. That package would not have to be coupled to MinVer in any way. It could be a general package for use anywhere, and would fit the philosophy of lego parts which I'm following with things like Bullseye, SimpleExec, and MinVer.

Thoughts?

bording:
I personally think it makes more sense to just add it to MinVer. Those attributes will only meaningful data if you know they've been set with the data by the versioning tool you're using. Each versioning tool also seems to have it's own way of populating `InformationalVersion`, so I'm not sure you could just expect to generally parse that information.

adamralph:
My third party library proposal is predicated on SemVer. MinVer will always put a SemVer version into `InformationalVersion`. In fact, you can do this today with https://www.nuget.org/packages/Semver/:

```C#
var version = SemVersion.Parse(
    assembly.GetCustomAttribute<AssemblyInformationalVersionAttribute>().InformationalVersion);
```

One problem I missed, however, is that, while this would probably be OK for apps, any library (like NServiceBus) would have take a dependency on https://www.nuget.org/packages/Semver/.

It's also worth bearing in mind that the KISS code to do this is quite trivial. Although this code looks a little clunky, it will _always_ work for the version produced by MinVer. It will only throw an exception when the version is _not_ SemVer compliant:

```C#
public class Version
{
    public int Major { get; set; }
    public int Minor { get; set; }
    public int Patch { get; set; }
    public string Prerelease { get; set; }
    public string BuildMetadata { get; set; }

    public static Version Parse(string version)
    {
        var versionAndMeta = version.Split(new[] { '+' }, 2);
        var numbersAndPre = versionAndMeta[0].Split(new[] { '-' }, 2);
        var numbers = numbersAndPre[0].Split('.');

        return new Version
        {
            Major = int.Parse(numbers[0]),
            Minor = int.Parse(numbers[1]),
            Patch = int.Parse(numbers[2]),
            Prerelease = numbersAndPre.Length > 1 ? numbersAndPre[1] : default,
            BuildMetadata = versionAndMeta.Length > 1 ? versionAndMeta[1] : default,
        };
    }
}
```

My reservations around adding this to MinVer are the following:

- It's language dependent. As I said, I know there are already some F# projects using SemVer. There could even be some VB ones. I really don't want to maintain this for several languages.
- It introduces more build target ordering considerations. For example, the readme describes [how to add targets to customise the package, assembly, or file versions](https://github.com/adamralph/minver#can-i-use-the-version-calculated-by-minver-for-other-purposes). We would have to guarantee that the target that injects this source file runs _after_ any targets like these added by users, or we'd have to add some explanation as to how to achieve that, which complicates the UX.
- We're duplicating what's already in `InformationalVersion`. This feels like the wrong direction from which to attack this problem. It seems natural for `InformationalVersion` to be the source of this information.

The cost of adding this to MinVer seems to be higher than the cost of library authors carrying around the `Version` class I show above.

bording:
> it's also worth bearing in mind that the KISS code to do this is quite trivial. Although this code looks a little clunky, it will always work for the version produced by MinVer. It will only throw an exception when the version is not SemVer compliant:

Sure, but I don't see why it makes sense to "force" a consumer to write that code vs. just giving them the information like this PR does.

> It's language dependent. As I said, I know there are already some F# projects using SemVer. There could even be some VB ones. I really don't want to maintain this for several languages.

It could start out as a C#-only feature, and only worry about additional language support if it's requested.

> It introduces more build target ordering considerations. For example, the readme describes how to add targets to customise the package, assembly, or file versions. We would have to guarantee that the target that injects this source file runs after any targets like these added by users, or we'd have to add some explanation as to how to achieve that, which complicates the UX.

The class introduced here is just reporting the values as calculated by MinVer. If someone decides to add custom targets to further alter things, and wants to see their alterations in the class, it seems like it would be up to them to ensure it runs after `MinVer` but before `MinVer_WriteVersionInformationFile`. The example targets in the readme would need a `BeforeTargets="MinVer_WriteVersionInformationFile"` added.

> We're duplicating what's already in InformationalVersion. This feels like the wrong direction from which to attack this problem. It seems natural for InformationalVersion to be the source of this information.

Nothing's being duplicated. It's just reporting the actual MSBuild properties that MinVer is calculating.

---
I think this would add value, and is adding some feature parity when compared to other versioning tools.

However, if you're still not convinced, then go ahead and close the PR! 😄

adamralph:
> Sure, but I don't see why it makes sense to "force" a consumer to write that code vs. just giving them the information like this PR does.

It wouldn't be "forcing" anyone to write anything. It's a question of where the solution belongs, whether that's in the consuming code, or in some package.

> Nothing's being duplicated. It's just reporting the actual MSBuild properties that MinVer is calculating.

It's _logical_ duplication. All the information is already available in `InformationalVersion`.

---

That said, you're doing a good job of persuading me. 😉

I'm still torn as to whether to include this. I appreciate the convenience, but somehow it doesn't feel right to introduce this here, in a language specific manner, which complicates the targets, when the information is already sitting there at runtime, in a language agnostic manner, in the form of `InformationalVersion`.

Another thing to consider is that some consumers may only need to perform very simple logic on `InformationalVersion`. E.g. in the case of NServiceBus, it's a one liner:

```c#
assembly.GetCustomAttribute<AssemblyInformationalVersionAttribute>().InformationalVersion
    .Split('-')[0]
```

---

BTW sorry that this conversation is dragging on, but I'm not seeing any urgency to this, so I'd rather thrash out all the pros and cons before committing to it.

bording:
> Another thing to consider is that some consumers may only need to perform very simple logic on InformationalVersion. E.g. in the case of NServiceBus, it's a one liner

I look at that reflection-based code example, and I don't see that as a very discoverable or desirable bit of code to have to write. It also means I'm having to take something that MinVer has provided separately to MSBuild and the parse it back out if I only needed a specific part.

It seems much simpler to me for my versioning tool to provide me directly with an easy way to get to all of the version bits it calculated, giving me the freedom to combine them however I want.

adamralph:
Discussed this with @bording earlier and we agreed to close this for now. If/when Particular explore the use of MinVer for NServiceBus then we can reconsider.

Issue #225, opened 2019-04-28
height not incremented as expected (or maybe my expectations are wrong)


mauroservienti:
This PR https://github.com/ServiceComposer/ServiceComposer.AspNetCore/pull/18 correctly generates a package versioned as `*-PR.18.alpha.0.*`, so far so good. I was expecting the `height` to increase with each commit, so `*-PR.18.alpha.0.0`, `*-PR.18.alpha.0.1`, `*-PR.18.alpha.0.etc..`, and if my memory works correctly this was the behavior I was observing. First question:

Is my expectation correct?

If this is the case. Due to reasons I force pushed on the PR, clearly visible on GitHub. Since then it seems that on the PR the MinVer generated version is stuck at `*-PR.18.alpha.0.2`, as if the height from the last tag is always 2.

Full build history is available here: https://ci.appveyor.com/project/ServiceComposer/servicecomposer-aspnetcore/history

Am I doing something wrong?

adamralph:
@mauroservienti I checked out the branch and it built `0.0.7-alpha.0.6.nupkg` as expected:

![image](https://user-images.githubusercontent.com/677704/56860315-a6bda200-6995-11e9-9f22-36b0a2e7c405.png)

mauroservienti:
Correct, however on the build server the behavior doesn’t match the local one. https://ci.appveyor.com/project/ServiceComposer/servicecomposer-aspnetcore Look at the last build I just kicked off. It’s still 2 as height, despite rebuilding the PR.

mauroservienti:
Is there any kind of logging I can enable on the build server to better understand what’s going on? Is it enough to enable msbuild verbose logging?

adamralph:
@mauroservienti I know what's happening. The CI build is not building the PR branch. It is building the _test merge_ of the PR branch to the master branch. Since master is already 1 commit ahead of the last tag, that test merge will always be 2 commits ahead of the tag.

mauroservienti:
OK, got it. Thanks @adamralph. Is this the expected behavior or should I expect also CI PR to behave like if I was building a branch? I personally expect the latter, at least this is what I was getting with GitVersion.

adamralph:
@mauroservienti yes, this is the expected behaviour. I suspect GitVersion does some kind of branch-based logic, and anything of that sort is deliberately avoided in MinVer.

[...]

PR #226, opened 2019-04-28
add var to identify the build to the PR example

adamralph:
Kudos to @mauroservienti for raising https://github.com/adamralph/minver/issues/225 which led to this.


mauroservienti:
thank you!

PR #250, opened 2019-08-03
add back to FAQ links

adamralph:


adamralph:
The browser's back button already does this. 🤦‍♂

PR #338, opened 2020-04-30
Added instructions for github actions that uses shallow clones

rvdkooy:
Hi Ralph,

While I was moving to github actions I found out that they also use shallow clones.

Only they use a depth of 1!
https://github.com/actions/checkout

In this pull request I updated the readme to mention this with the link to the docs on how to fix this.

I Hope this could be helpfull for others.

btw: Thanks for this great package!

R


adamralph:
@rvdkooy thanks for raising this. It looks like this is [new in actions/checkout 2.0.0](actions/checkout):

> - Improved fetch performance
>   - The default behavior now fetches only the commit being checked-out

Another reminder that you should always pin your version. 🙂

PR #344, opened 2020-05-25
Add more BeforeTargets to MinVer target

bording:
https://github.com/dotnet/sdk/pull/10613 made a change to the `GenerateAssemblyInfo` target to fix  a [problem](https://github.com/dotnet/sdk/issues/10614) related to Source Link.

This change ensures that the `MinVer` target is called early enough to get the version values populated before `GenerateAssemblyInfo` writes them to disk in the generated `AssemblyInfo.cs` file.

Fixes #343

Issue #343, opened 2020-05-25
Assembly attributes not written (or overwritten) using SDK 3.1.300


joacar:
It works great if built with SDK < 3.1.300 but fails on 3.1.300.

## Version(s)

2.2.0

## To reproduce

There is a tag 2.1.3 and that value should be displayed at first run.

1. Run included console app
1. Change global.json to 3.1.300
1. Run included console app

[ConsoleApp2.zip](https://github.com/adamralph/minver/files/4678078/ConsoleApp2.zip)

## Expected behavior

See informational version 2.1.3 as calculated by MinVer

## Actual behaviour

Default 1.0.0

## Workarounds

Haven't find any. Tried to elaborate will target order but no luck.

## Additional context

Only seen this in 3.1.300

bording:
@joacar See #344 for more details and the fix.

joacar:
@bording Great job with a quick fix!

adamralph:
@joacar thanks for raising this and for the detailed repro! Since this was caused by a breaking change in SDK 3.1.300, I'm not treating this as a bug in MinVer, so I'm withdrawing this issue in favour of https://github.com/adamralph/minver/pull/347.

Issue #342, opened 2020-05-21
Error (dotnet 3.1.300; MinVer 2.2): The specified version string does not conform to the recommended format - major.minor.build.revision


ghorsey:
## Version(s)

MinVer: `v2.2`
dotnet: `v3.1.300`
```
$ dotnet --info
.NET Core SDK (reflecting any global.json):
 Version:   3.1.300
 Commit:    b2475c1295

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.18363
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\3.1.300\

Host (useful for support):
  Version: 3.1.4
  Commit:  0c2e69caa6

.NET Core SDKs installed:
  2.2.207 [C:\Program Files\dotnet\sdk]
  3.1.201 [C:\Program Files\dotnet\sdk]
  3.1.202 [C:\Program Files\dotnet\sdk]
  3.1.300 [C:\Program Files\dotnet\sdk]

.NET Core runtimes installed:
  Microsoft.AspNetCore.All 2.1.18 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.All]
  Microsoft.AspNetCore.All 2.2.8 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.All]
  Microsoft.AspNetCore.App 2.1.18 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 2.2.8 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 3.1.3 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 3.1.4 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.NETCore.App 2.1.18 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 2.2.8 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 3.1.3 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 3.1.4 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
[...]

PR #347, opened 2020-05-26
Support .NET Core SDK 3.1.300 and later

adamralph:
.NET Core SDK 3.1.300 contains a [breaking change](https://github.com/dotnet/sdk/issues/10614) which results in projects versioned with MinVer producing assemblies which do not have the correct `AssemblyFileVersion`, `AssemblyInformationalVersion`, and `AssemblyVersion`. This effectively means that MinVer 2.2.0 and earlier do not support .NET Core SDK 3.1.300.

This enhancement adds support for .NET Core SDK 3.1.300 and later.

Thanks to @ghorsey and @joacar for spotting this and raising #342 and #343 respectively, and thanks to @bording for fixing this in https://github.com/adamralph/minver/pull/344, which was cherry-picked into this branch.

### DIFF
--- a/targets/Program.cs
+++ b/targets/Program.cs
@@ -57,7 +57,7 @@ internal static class Program
                     Path.Combine(testProject, "global.json"),
 @"{
   ""sdk"": {
-    ""version"": ""3.1.102""
+    ""version"": ""3.1.300""
   }
 }
 "
--- a/targets/Program.cs
+++ b/targets/Program.cs
@@ -50,6 +50,16 @@ internal static class Program
                 var source = Path.GetFullPath("./MinVer/bin/Release/");
                 var version = Path.GetFileNameWithoutExtension(Directory.EnumerateFiles(source, "*.nupkg").First()).Split("MinVer.", 2)[1];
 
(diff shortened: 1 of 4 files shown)


targets/Program.cs
```diff
@@ -49,6 +53,16 @@ public static async Task Main(string[] args)
...
 
+                File.WriteAllText(
+                    Path.Combine(testProject, "global.json"),
+@"{
+  ""sdk"": {
+    ""version"": ""3.1.300""
+  }
+}
+"
+                    );
```
thomaslevesque, line 64: I'm late to the party, but here's a simpler way:
```
dotnet new globaljson --sdk-version 3.1.300 --force
```

adamralph: Thanks @thomaslevesque!

BTW why would I need the `--force` option?

adamralph: FYI — https://github.com/adamralph/minver/pull/350

thomaslevesque: To overwrite the existing file, like your current code is doing

PR #350, opened 2020-05-27
simpler global json creation for package tests

adamralph:
Thanks to @thomaslevesque for [suggesting this](https://github.com/adamralph/minver/pull/347#discussion_r430715600)!

### DIFF
--- a/targets/Program.cs
+++ b/targets/Program.cs
@@ -53,16 +53,7 @@ internal static class Program
                 var source = Path.GetFullPath("./MinVer/bin/Release/");
                 var version = Path.GetFileNameWithoutExtension(Directory.EnumerateFiles(source, "*.nupkg").First()).Split("MinVer.", 2)[1];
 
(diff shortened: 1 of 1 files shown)


targets/Program.cs
```diff
@@ -53,16 +53,7 @@ public static async Task Main(string[] args)
...
-                    Path.Combine(testProject, "global.json"),
-@"{
-  ""sdk"": {
-    ""version"": ""3.1.300""
-  }
-}
-"
-                    );
-
+                await RunAsync("dotnet", "new globaljson --sdk-version 3.1.300", testProject);
```
thomaslevesque, line 56: Without the --force option, it will fail if the file already exists. The current code just overwrites the file

adamralph: @thomaslevesque thanks for clarifying. Given that the directory is already ensured to be empty on line 51, we can do without the `--force` option.

PR #366, opened 2020-08-11
CLI output has new line terminator(s)

adamralph:
Can cause problems when parsing the output. For example, using the `semver` package:

```c#
var versionString = SimpleExec.Command.Read("minver");
var version = Semver.SemVersion.Parse(versionString);
```

This results in an "Invalid version" exception.

The version needs to be trimmed first:

```c#
var versionString = SimpleExec.Command.Read("minver").Trim();
var version = Semver.SemVersion.Parse(versionString);
```

### DIFF
--- a/minver-cli/Program.cs
+++ b/minver-cli/Program.cs
@@ -62,7 +62,7 @@ namespace MinVer
                 var version = Versioner.GetVersion(workDir, tagPrefixOption.Value(), minMajorMinor, buildMetaOption.Value(), autoIncrement, defaultPreReleasePhaseOption.Value(), log);
 #endif
 
-                    Console.Out.WriteLine(version);
+                    Console.Out.Write(version);
 
                     return 0;
                 });


adamralph:
https://pubs.opengroup.org/onlinepubs/009695399/basedefs/xbd_chap03.html#tag_03_392

> *3.392 Text File*
> 
> A file that contains characters organized into one or more lines. The lines do not contain NUL characters and none can exceed {LINE_MAX} bytes in length, including the <newline>. Although IEEE Std 1003.1-2001 does not distinguish between text files and binary files (see the ISO C standard), many utilities only produce predictable or meaningful output when operating on text files. The standard utilities that have such restrictions always specify "text files" in their STDIN or INPUT FILES sections.

https://pubs.opengroup.org/onlinepubs/009695399/basedefs/xbd_chap03.html#tag_03_205

> *3.205 Line*
> 
> A sequence of zero or more non- <newline>s plus a terminating <newline>.

Given that `stdout` is a special _file_, it should contain _lines_ and, in the case of `minver-cli`, the output is a single _line_.

Issue #363, opened 2020-07-30
Problem with subtree


karlika:
We have a project where we added sources from another repo using git subtree:
`git subtree add –-prefix src/mainframe https://[REPOPATH]/mainframe.git master --squash`

If I now run the minver CLI on this repo it ignores a whole lot of commits and computes a depth which is too low.
Actually in the current run (3) it was lower than the previous run before the last subtree change (4).

I'm not a 100% sure, how my colleagues did the subtree update, but it looks like they manually removed the folder where the subtree was added, committed and did the `git subtree add` again.

Is there any way we can get a valid depth with the subtree?

[minver_trace_log.txt](https://github.com/adamralph/minver/files/4995909/minver_trace_log.txt)

adamralph:
Hi @karlika, thanks for raising this.

I'm afraid I have no idea about subtrees and how they work. I've never used one. I would have no idea where to start in diagnosing the problem.

As you can see from the log you attached, MinVer simples relies on the output from the various git commands it is running. If you can figure out what is required to support subtrees then I may accept a pull request to add the feature, if it does not introduce too much complexity.

PR #367, opened 2020-08-19
Ignore 'root commits' created by subtree add 

karlika:
Using the fact that 'git subtree add' adds a comment to the commit containing 'git-subtree-dir' we try to identify such commits and thus avoid that they are erroneously identified as root commit.
#363

### DIFF
--- a/MinVer.Lib/Commit.cs
+++ b/MinVer.Lib/Commit.cs
@@ -10,7 +10,8 @@ namespace MinVer.Lib
 
         public string ShortSha => this.Sha.Substring(0, 7);
 
-        public bool SubtreeRoot { get; } = false;
+        public bool SubtreeRoot { get; }
+
 
         public List<Commit> Parents { get; } = new List<Commit>();
     }
--- a/MinVer.Lib/Commit.cs
+++ b/MinVer.Lib/Commit.cs
@@ -4,11 +4,13 @@ namespace MinVer.Lib
 
     internal class Commit
     {
-        public Commit(string sha) => this.Sha = sha;
+        public Commit(string sha, bool subtreeRoot = false) => (this.Sha, this.SubtreeRoot) = (sha, subtreeRoot);
 
         public string Sha { get; }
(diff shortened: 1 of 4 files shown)


MinVer.Lib/Commit.cs
```diff
@@ -4,12 +4,14 @@ namespace MinVer.Lib
...
     internal class Commit
     {
-        public Commit(string sha) => this.Sha = sha;
+        public Commit(string sha, bool subtreeRoot = false) => (this.Sha, this.SubtreeRoot) = (sha, subtreeRoot);
 
         public string Sha { get; }
 
         public string ShortSha => this.Sha.Substring(0, 7);
 
+        public bool SubtreeRoot { get; } = false;
```
sungam3r, line 13:
```suggestion
        public bool SubtreeRoot { get; }

```

adamralph:
@karlika @sungam3r thanks for this. I'll take a closer look ASAP but it won't be until some time next week.

adamralph:
@karlika I haven't forgotten about this. I started to look into it but I got stuck fairly quickly. The problem is that I know next to nothing about subtrees.

As a first step, I'd like to reproduce the current scenario, i.e. the wrong version being produced. Would you be able to provide me with the steps required to do that? Ultimately I'd like to get that into a test somehow.

karlika:
@adamralph Actually I'm not sure if you could create a test case for that, because you need a remote repo for the sbutree add command.
I created a public repo where I already added a subtree here: https://github.com/karlika/subtreetest.git
The subtree was added using this command:
`git subtree add -P dummy https://github.com/karlika/dummy.git master`
The depth reported by minver is 2 but it should actually be 3.

adamralph:
Thank you @karlika, that was a great help.

It turns out that this is a bug. Fortunately, [the fix is very simple](https://github.com/adamralph/minver/pull/377).

adamralph:
closing favour of #377

PR #377, opened 2020-09-14
first root commit is not being used

adamralph:
## Version(s)

`2.3.0`

## To reproduce

image

Run MinVer against a repo which looks like ☝️.

## Expected behaviour

The root commit on the first (black) branch is used, since that branch was the current branch when the merged was performed. The version is `0.0.0-alpha.3`.

## Actual behaviour

The root commit on the second (blue) branch is used. The version is `0.0.0-alpha.2`.

## Workarounds

Tag some commit to coerce the version into being something you expect.

## Additional context

Thanks to @karlika for [spotting this](https://github.com/adamralph/minver/issues/363)!

PR #378, opened 2020-09-14
first tag with a given version is not used

adamralph:
## Version(s)

`2.3.0`

## To reproduce

image

Run MinVer against a repo which looks like ☝️.

## Expected behaviour

The tag on the first (blue) branch is used, since that branch was the current branch when the merged was performed. The version is `1.0.0-alpha.2`.

## Actual behaviour

The tag on the second (yellow) branch is used. The version is `1.0.0-alpha.1`.

## Workarounds

Tag some commit to coerce the version into being something you expect.

## Additional context

Found while fixing #377. Thanks again to @karlika for [spotting that bug](https://github.com/adamralph/minver/issues/363)!

Another approach is to ignore the `1.0.00` tag, since it is not valid SemVer due to the leading zero in the patch version. However, there may be other cases where the two tags could produce the same version, e.g. using special chars, so it's probably better to just assume that it is possible for two tags to produce the same version.

### DIFF
--- a/MinVer.Lib/Repository.cs
+++ b/MinVer.Lib/Repository.cs
@@ -54,7 +54,6 @@ namespace MinVer.Lib
             var commitsChecked = new HashSet<string>();
             var count = 0;
             var height = 0;
-            var foundRootCommit = false;
             var candidates = new List<Candidate>();
             var commitsToCheck = new Stack<(Commit, int, Commit)>();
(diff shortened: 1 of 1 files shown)


adamralph:
release in [2.3.1-beta.2](https://www.nuget.org/packages/MinVer/2.3.1-beta.2)

PR #384, opened 2020-09-28
add global.json

adamralph:


.github/workflows/ci.yml
```diff
@@ -14,12 +14,13 @@ jobs:
       fail-fast: false
       matrix:
         runs-on: [macos-10.15, ubuntu-16.04, ubuntu-18.04, windows-2016, windows-2019]
-    name: ${{ matrix.runs-on }}
+        dotnet: ['3.1.100', '3.1.402']
```
bording, line 17: Since these are SDK versions and not actual .NET Core versions, what do you gain from this?

adamralph: @bording the intention is to ensure that the added `global.json` is telling the truth. I.e. you can build the solution with 3.1.100. It turns out that's not true though, because we pinned the tests to a specific SDK in #347.

bording: If you are controlling the SDK version in the CI script, why do you care about being able to build the solution with multiple SDKs?

Also, seems like `"rollForward": "latestMajor"` could have some unintended implications. The GitHub-hosted runners have many versions of the SDK installed, so it seems like as soon as they update, the latest one would be chosen over the one you've specified via the `setup-dotnet` action.

adamralph: > ...why do you care about being able to build the solution with multiple SDKs?

I have a pet hate for projects which pin an exact version of the SDK. It's usually OK if it's the latest, but if not then there's little I hate more than a project refusing to build just because I have, for example, 3.1.403, and I happened to skip 3.1.402.

It's one thing to pin the SDK version for CI, which is desirable, but that same restriction should not apply to local development, IMO.

Having said that, while I believe the above applies for most of my projects, MinVer is a little special in this regard. As we saw in https://github.com/adamralph/minver/pull/347, MinVer is very sensitive to SDK releases. 3.1.300 was effectively a breaking release for MinVer. So I'm now reevaluating my policy on this wrt MinVer. As things currently stand in `master`, it is not possible to run the build script without a specific version of the SDK (3.1.402) due to the `global.json` used in the package tests, and maybe it has to stay that way, which means we should also add a `global.json` to the repo, pinned to the same version (or perhaps with roll-forward).

> Also, seems like "rollForward": "latestMajor" could have some unintended implications. The GitHub-hosted runners have many versions of the SDK installed, so it seems like as soon as they update, the latest one would be chosen over the one you've specified via the setup-dotnet action.

The `setup-dotnet` action ensures that the specified version is used by subsequent steps. More information on that in [this conversation](https://github.com/dependabot/dependabot-core/issues/2442#issuecomment-699620258).

bording: > I have a pet hate for projects which pin an exact version of the SDK. 

Right there with you on that.


> Having said that, while I believe the above applies for most of my projects, MinVer is a little special in this regard. As we saw in #347, MinVer is very sensitive to SDK releases.

The distinction here is that it can be sensitive to SDK version when it is being consumed and used, not when it's being built itself, which I'm sure you're aware of!

> As things currently stand in master, it is not possible to run the build script without a specific version of the SDK (3.1.402) due to the global.json used in the package tests,

At this point, what is the benefit of having that globl.json created in the tests? If you're trying to catch bugs with new SDK releases, then locking to an older one in the tests doesn't do that. 


> The setup-dotnet action ensures that the specified version is used by subsequent steps. More information on that in this conversation.

I was aware that it would make the selected version the one on the path, so the "default" one, but I wasn't aware that it would also hide all other SDK versions from being seen at all. At this point I'd like to see the `dotnet --info` output to confirm what is discoverable.

If it hiding other versions, then that seems like it could be problematic as well if it's also hiding the runtimes installed by the older SDKs. Seems like you could get unanticipated runtime roll forward that way.

adamralph: > The distinction here is that it can be sensitive to SDK version when it is being consumed and used, not when it's being built itself, which I'm sure you're aware of!

That's true, but it means that it has to be _tested_ against specific versions of the SDK during the build. Specifically, there was a break in 3.1.300, so we need to ensure it is tested against versions later than that. If we allow 3.1.300 for local dev, then theoretically a change can be made locally which doesn't work in CI. I don't think this is much of a problem though.

> At this point, what is the benefit of having that globl.json created in the tests? If you're trying to catch bugs with new SDK releases, then locking to an older one in the tests doesn't do that.

The idea there was to ensure that the SDK being used is >= 3.1.300 to ensure the break in that version was catered for. However, given that `setup-dotnet` guarantees the vesion, it's actually redundant in CI. I'm still unsure how to handle the local dev situation, however.

> At this point I'd like to see the `dotnet --info` output to confirm what is discoverable.

That is run on every build, e.g. https://github.com/adamralph/minver/pull/384/checks?check_run_id=1178364565#step:3:1. As you can see, only the single SDK version installed by `setup-dotnet` is available.

> If it hiding other versions, then that seems like it could be problematic as well if it's also hiding the runtimes installed by the older SDKs. Seems like you could get unanticipated runtime roll forward that way.

That's fine. I don't see any reason to test MinVer on older runtimes.

bording: > That is run on every build, e.g. https://github.com/adamralph/minver/pull/384/checks?check_run_id=1178364565#step:3:1. As you can see, only the single SDK version installed by setup-dotnet is available.

That is interesting that it can hide all the other installed SDKs. I assume that if you removed the `setup-dotnet` action, then you would be able to see all of them.


> The idea there was to ensure that the SDK being used is >= 3.1.300 to ensure the break in that version was catered for. 

If that's the goal, then it seems like you could accomplish that by:
1. bump the version in this PR's global.json file to 3.1.300
2. Remove the changes in this PR to ci.yml
3. Change the global.json file created for the tests to look like the one is this PR.

That would ensure that local development and CI both had at least 3.1.300, but would allow newer versions to be used as well without restriction.

adamralph: > 1. bump the version in this PR's global.json file to 3.1.300
> 2. Remove the changes in this PR to ci.yml
> 3. Change the global.json file created for the tests to look like the one is this PR.

👍 that makes sense. I did consider bumping the minimum to 3.1.300, but for some reason I got hung up on allowing people to build with < 3.1.300. I guess that doesn't really make sense though.

There's just one wrinkle: I'd still like to _test_ against older SDK's. In fact, the README says that MinVer works with SDK 2.1.300 or later, but that isn't verified anywhere. Anyway, that's outside the scope of this PR.

I'll go with your suggestion for now, thanks! But with a slight modification: if the version in the test `global.json` is going to be older than that in the repo `global.json`, it may as well be 2.1.300, to reflect what the README is saying. Actually testing against that version can be figured out in another PR.

PR #398, opened 2020-10-17
test on dotnet 2.1.300

adamralph:


adamralph:
Running the entire build on a 2.1 SDK isn't going to fly. I got `dotnet build` working locally but then I received 666 👿 warnings about incompatible analysers, etc. etc. etc. Will find another way.

PR #422, opened 2020-11-22
Less test output

adamralph:


adamralph:
"read Git in tests instead of running it to suppress output" doesn't make any difference.

adamralph:
not sure we want to build the test project with quiet verbosity, since ultimately, we want MinVer verbosity to inherit from that

PR #474, opened 2021-01-07
Package on build

bording:
Alternative approach to #473

### DIFF
--- a/MinVer/MinVer.csproj
+++ b/MinVer/MinVer.csproj
@@ -1,11 +1,13 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
     <AnalysisMode>AllEnabledByDefault</AnalysisMode>
+    <CopyLocalLockFileAssemblies>true</CopyLocalLockFileAssemblies>
     <DefineConstants>$(DefineConstants);MINVER</DefineConstants>
     <Description>Minimalist versioning for .NET SDK-style projects using Git tags.</Description>
     <DevelopmentDependency>true</DevelopmentDependency>
     <EnableNETAnalyzers>true</EnableNETAnalyzers>
+    <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
     <IncludeBuildOutput>false</IncludeBuildOutput>
     <NoWarn>$(NoWarn);NU5100;NU5105</NoWarn>
     <OutputType>Exe</OutputType>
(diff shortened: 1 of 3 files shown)


MinVer/MinVer.csproj
```diff
@@ -40,7 +42,12 @@
     <None Remove="buildMultiTargeting\**\*" />
     <Content Include="build\**\*" PackagePath="build" />
     <Content Include="buildMultiTargeting\**\*" PackagePath="buildMultiTargeting" />
-    <Content Include="$(OutputPath)\publish\**\*" PackagePath="minver" Visible="false" Link="hidden" />
   </ItemGroup>
 
+  <Target Name="AddMinVerOutput" BeforeTargets="_GetPackageFiles">
```
adamralph, line 47:
```text
BeforeTargets="_GetPackageFiles"
```

This is the key thing. I think I'll never understand how you find these things! 😆

adamralph (approved):
Wonderful! Thanks very much @bording! 😍

MinVer/MinVer.csproj
bording: Strictly speaking that target just needs to run after all the artifacts are compiled and written to disk, so I could have used something like `AfterTargets="CoreCompile"` instead. However, since `_GetPackageFiles` is the actual target that does the work of turning the various build action item groups into the list of files that need to be packaged, I decided to use that one instead.

adamralph: > since `_GetPackageFiles` is the actual target that does the work of turning the various build action item groups into the list of files that need to be packaged

How did you find that out?

bording: > How did you find that out?

The key tools for any sort of MSBuild investigation is to get a binary log, which is easy to do in VS with the [Project System Tools](https://marketplace.visualstudio.com/items?itemName=VisualStudioProductTeam.ProjectSystemTools) extension), and then analyze that log file with the [MSBuild Structured Log Viewer](https://msbuildlog.com/).

From there, it's possible to see the actual targets being run and actually open up the targets files and look at what they are doing.

I've spent a lot of time looking at the actual NuGet packaging targets files in the past, so I was already aware that there was a target that did the conversion, but I did have to look at the log file to rediscover the name of it.

I'd be happy to walk you through my process for this sort of thing if you're interested.

adamralph: Thanks, I grabbed Project System Tools (and updated my installation of MSBuild Structured Log Viewer) and I managed to examine the build log for the MinVer project.

Still, there is a _whole_ lot going on between `CoreCompile` and `_GetPackageFiles`, and I find the `targets` source files very hard to read. I'm not sure that my workflow for identifying where to inject my target would ever be much more than trial and error. 🤷

image

PR #547, opened 2021-06-17
cache results to avoid multiple version calculations for the same inputs

adamralph:


adamralph:
@bording I thought I'd take a look a this, with reference to this [comment](https://github.com/dotnet/sourcelink/issues/122#issuecomment-621475891) that you pointed out to me, but it looks like [`MSBuild.UseResultsCache`](https://docs.microsoft.com/en-us/dotnet/api/microsoft.build.tasks.msbuild.useresultscache) has been deprecated and now "does nothing", so I guess I'd have to implement this caching myself.

adamralph:
FTR there's also [`CallTarget.UseResultsCache`](https://docs.microsoft.com/en-us/dotnet/api/microsoft.build.tasks.calltarget.useresultscache), but it has also been deprecated and "does nothing".

bording:
> @bording I thought I'd take a look a this, with reference to this [comment](https://github.com/dotnet/sourcelink/issues/122#issuecomment-621475891) that you pointed out to me, but it looks like [`MSBuild.UseResultsCache`](https://docs.microsoft.com/en-us/dotnet/api/microsoft.build.tasks.msbuild.useresultscache) has been deprecated and now "does nothing", so I guess I'd have to implement this caching myself.

Looking through the MSBuild codebase, it does seem to confirm that `UseResultsCache` is just a stub in the public API now, so that approach does appear to be a dead end.

The `RegisterTaskObject`/`GetRegisteredTaskObject` looks like it would still work, but I believe you'd have to be running the versioning logic from an MSBuild task to use those.


Assuming you aren't interested in introducing an MSBuild task assembly, then it does seem like you'd have to build your own caching logic. It seems like it would need to be some sort of disk-based results file.

adamralph:
Thanks for the input @bording. I'm going to close this as a `wontfix`. FWIW I had another look at https://github.com/adamralph/minver/issues/112 and I still don't think there is enough ROI on switching to an MSBuild task, even taking this feature into consideration.

PR #560, opened 2021-06-20
restrict MultipleProjects tests to SDK 5 and later

adamralph:


MinVerTests.Packages/Net5PlusFact.cs
```diff
@@ -0,0 +1,27 @@
...
+            get => base.Skip ??
+                (
+                    (
+                        Sdk.Version != null &&
+                        (
+                            Sdk.Version.StartsWith("2.", StringComparison.Ordinal) ||
+                            Sdk.Version.StartsWith("3.", StringComparison.Ordinal)
+                        )
+                    )
+                    ? "Not .NET 5 or later"
```
sungam3r, line 20: "Not .NET 5 or later" ? Does the meaning of the attribute not in the opposite?

adamralph: If the version starts with `2.` or `3.` then it is _not_ .NET 5 or later.

Issue #362, opened 2020-07-09
Get more version variables from CLI


chadly:
## Use case

I'm using MinVer to successfully version my .NET projects and it works exactly as I would like it to. I'm trying to extend my usage of it to create versioned docker images of my .NET app.

As an example, my .NET app successfully gets versioned to v1.2.3. I would then like to use the minver CLI to get that same version information to version the docker image. As it is now, I can get the string `1.2.3` out of the minver CLI. But I would also like to get these variables that get passed to MSBuild:

> MinVer sets the following custom properties:
> * `MinVerVersion`
> * `MinVerMajor`
> * `MinVerMinor`
> * `MinVerPatch`
> * `MinVerPreRelease`
> * `MinVerBuildMetadata`

I would like to use this information to tag the docker image using the major, minor, patch convention: e.g. `v1`, `v1.2`, and `v1.2.3` so that someone consuming the docker image can depend on the `v1` tag and get all minor/patch updates.

## Description

I'm not sure how it would work. Maybe there can be another command line flag `--environment-variables` or something to set all those version numbers as environment variables which I would then be able to use in the rest of my build script.

adamralph:
@chadly thanks for raising this.

I've been asked this before, and the answer I favour is to deconstruct the version on the consuming side. This can be done reliably because the version format is deterministic. I don't know which language you are using to write your build script, but in C# this is fairly easy:

```C#
var metaSplit = version.Split('+', 2);
var preSplit = metaSplit[0].Split('-', 2);
var rtmSplit = preSplit[0].Split(".");
var (major, minor, patch, preRelease, buildMetadata) =
    (rtmSplit[0], rtmSplit[1], rtmSplit[2], preSplit.ElementAtOrDefault(1), metaSplit.ElementAtOrDefault(1));
```

Returning separate variables from the CLI would require the introduction a serialisation format and a command line option to switch it on, and I'd like to avoid that added complexity if possible.

Also, AFAIK, setting environment variables will not work, since env vars set by a child process are only scoped to that process, and not the calling process.

chadly:
I'm using GitHub Actions with [this action to run MinVer](https://github.com/thefringeninja/action-minver). Now that you mention it, it does seem like this would be a better option to add to the GH Action since that can easily define more than one output and can parse the outputted MinVer version using something like the `semver` npm package.

Thanks for the feedback!

adamralph:
> this would be a better option to add to the GH Action

👍 nice idea.

PR #621, opened 2021-11-03
feature: Supports GitHub Action set-output

andersklepaker:
Hi! 

Thanks for the effort you and others have put into this prosject making versioning with git real simple! 

I've taken the liberty to create this PR to simplify using MinVer computed values in a GitHub Action workflow. This PR solves this by adding a switch `-s|--show-github-action-output` to the `minver-cli.exe` making it output the following during the workflow. 

`minver-cli.exe -s`
```
::set-output name=MinVerMajor::0
::set-output name=MinVerMinor::0
::set-output name=MinVerPatch::0
::set-output name=MinVerPreReleaseIdentifier::alpha.0
::set-output name=MinVerBuildMetaData::
::set-output name=MinVerPreHeight::0
::set-output name=MinVerVersion::0.0.0-alpha.0
```

This allows other steps in the workflow to access those variables, i.e:
```
MinVerVersion = ${{ steps.minver.outputs.MinVerVersion }}
MinVerMajor = ${{ steps.minver.outputs.MinVerMajor }}
MinVerMinor = ${{ steps.minver.outputs.MinVerMinor }}
MinVerPatch = ${{ steps.minver.outputs.MinVerPatch }}
MinVerPreReleaseIdentifier = ${{ steps.minver.outputs.MinVerPreReleaseIdentifier }}
MinVerBuildMetaData = ${{ steps.minver.outputs.MinVerBuildMetaData }}
MinVerPreHeight = ${{ steps.minver.outputs.MinVerPreHeight }}
MinVerVersion = ${{ steps.minver.outputs.MinVerVersion }}
```
Ref: https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#jobsjob_idoutputs

My usecase for doing this is to be able to tag a docker image at build time with the values from MinVer while not using the complete semver string returned. 

If this has already been resolved or there are better solutions to this problem please close this PR!

### DIFF
--- a/MinVer.Lib/Version.cs
+++ b/MinVer.Lib/Version.cs
@@ -7,27 +7,28 @@ namespace MinVer.Lib
 {
     public class Version : IComparable<Version>
     {
-        private readonly int major;
-        private readonly int minor;
-        private readonly int patch;
         private readonly List<string> preReleaseIdentifiers;
-        private readonly int height;
-        private readonly string buildMetadata;
+        public int Major { get; }
+        public int Minor { get; }
+        public int Patch { get; }
+        public int Height { get; }
+        public string BuildMetaData { get; }
+        public string PreReleaseIdentifier => this.preReleaseIdentifiers.Count == 0 ? "" : $"{string.Join(".", this.preReleaseIdentifiers)}";
 
         public Version(string defaultPreReleasePhase) : this(0, 0, 0, new List<string> { defaultPreReleasePhase, "0" }, 0, null) { }
 
         private Version(int major, int minor, int patch, IEnumerable<string> preReleaseIdentifiers, int height, string buildMetadata)
         {
-            this.major = major;
-            this.minor = minor;
-            this.patch = patch;
+            this.Major = major;
+            this.Minor = minor;
+            this.Patch = patch;
             this.preReleaseIdentifiers = preReleaseIdentifiers?.ToList() ?? new List<string>();
-            this.height = height;
-            this.buildMetadata = buildMetadata;
+            this.Height = height;
+            this.BuildMetaData = buildMetadata;
         }
 
         public override string ToString() =>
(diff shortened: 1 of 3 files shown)


adamralph:
Thanks for raising this @andersklepaker! I'm a little busy at the moment but I'll take a closer look ASAP.

adamralph:
@andersklepaker do you explicitly require height as a separate variable?

Currently height is not exposed publicly. It is only exposed as part of what makes up `MinVerPreRelease` (an output property from the `MinVer` package). It doesn't really have any meaning on it's own, and has no representation in SemVer. It's really only something that is used internally by MinVer to construct the pre-release identifiers making up the SemVer version.

Unless you have a specific use case for the height, I'd prefer not to expose it. If it were to be exposed, then we'd need another name for the "non-height" part of the pre-release identifiers, because the pre-release identifiers must, by SemVer definition, contain the height.

adamralph:
Just to explore other options:

@chadly wrt to https://github.com/adamralph/minver/issues/362#issuecomment-656718811 I see that the https://github.com/thefringeninja/action-minver/pull/6 has stagnated, but did you find another solution for exposing the version parts to your GHA workflow?

adamralph:
FWIW, I just came across https://github.com/marketplace/actions/action-minver-refreshed which outputs the following:

- `version`
- `major`
- `minor`
- `patch`
- `prerelease`

@andersklepaker would that suit your needs? It doesn't output the build metadata but I guess that could be added easily if required.

andersklepaker:
> FWIW, I just came across [github.com/marketplace/actions/action-minver-refreshed](https://github.com/marketplace/actions/action-minver-refreshed) which outputs the following:
> 
> * `version`
> * `major`
> * `minor`
> * `patch`
> * `prerelease`
> 
> @andersklepaker would that suit your needs? It doesn't output the build metadata but I guess that could be added easily if required.

Yes - this is would work. Thanks, and sorry for not finding this myself :)

adamralph:
@andersklepaker no worries. I'll close this for now, on that basis, but feel free to re-open or chime in again if required.

PR #655, opened 2022-01-01
add explicit default values to CLI help

adamralph:
building on https://github.com/adamralph/minver/pull/653 and https://github.com/natemcmaster/CommandLineUtils/issues/82

## Before

image

## After

image


adamralph:
Doesn't work, since the default value, instead of a `null`, prevents overriding env vars via the CLI.

PR #676, opened 2022-02-24
Fix dependency order with NuGet

### LINKED ISSUE #675: Invalid order with NuGet pack leading to invalid outputs
Hey, 
Thank you for MinVer. I'm starting to use it in all my projects and it simplifies a lot what I was doing manually! 👍 

## Version(s)

`2.5.0`

## Problem

I'm developing a tooling that simplifies the whole build/publish process for .NET libraries/apps on GitHub ([dotnet-releaser](https://github.com/xoofx/dotnet-releaser)) and I discovered a weird behavior while using MinVer on a multi-targeting version (It seems that it doesn't affect packages with only one target framework).

I'm calling msbuild directly with a special logger from which I can extract results from msbuild and I discovered that the version used to name output files in msbuild with MinVer when packing is wrong. 

For example, below, you have a screenshot of msbuild log:

image

It is supposed to create the package `Tomlyn.0.11.0.nupkg` and the pack command does it, but actually, when you look in the details, msbuild will think that it needs to generate `Tomlyn.1.0.0.nupkg` and will always re-run even if you build it just before.

Reason is that the file output is calculated in the target `_GetOutputItemsFromPack` which called by `_CalculateInputsOutputsForPack` (code [here](https://github.com/NuGet/NuGet.Client/blob/352ba94f120f1a3481a2cae6190a1947a85a160f/src/NuGet.Core/NuGet.Build.Tasks.Pack/NuGet.Build.Tasks.Pack.targets#L178-L194)) but MinVer runs only just before `GenerateNuspec` (as you can see in the screenshot above), but not before `_GetOutputItemsFromPack`.

## Expected behaviour

NuGet msbuild tasks should get the version calculated from MinVer in order to correctly compute output packages.

## Actual behaviour

NuGet is originally thinking that the version is `1.0.0` instead the version calculated by MinVer. If you run multiple times NuGet pack with MinVer, it will always have to repack the package thinking that it doesn't exist (because 1.0.0 is not created in the end)

## Workarounds

No great workaround as I would have to make my tooling MinVer aware to force MinVer to insert before `_GetOutputItemsFromPack`.

## Fix Suggestion

Add a msbuild dependency to `_GetOutputItemsFromPack` instead of only `GenerateNuSpec`. 
Hacking MinVer locally and it is now correctly generating the correct versions:

image

I will make a PR with this suggestion fix.

xoofx:
Fixes #675

### DIFF
--- a/MinVer/build/MinVer.targets
+++ b/MinVer/build/MinVer.targets
@@ -3,6 +3,7 @@
   <PropertyGroup>
     <MSBuildAllProjects Condition="'$(MSBuildAssemblyVersion)' == '' Or '$(MSBuildAssemblyVersion)' &lt; '16.0'">$(MSBuildAllProjects);$(MSBuildThisFileFullPath)</MSBuildAllProjects>
     <GetPackageVersionDependsOn>$(GetPackageVersionDependsOn);MinVer</GetPackageVersionDependsOn>
+    <GenerateNuspecDependsOn>$(GenerateNuspecDependsOn);MinVer</GenerateNuspecDependsOn>
   </PropertyGroup>
 
   <PropertyGroup>


xoofx:
Another fix would be to insert MinVer into `GenerateNuspecDependsOn`. I just saw that it was done on `GetPackageVersionDependsOn` but it looks like it could be also added to `GenerateNuspecDepends`?

What do you prefer? (depending on a "private" target `_GetOutputItemsFromPack` is probably not great)

xoofx:
> Another fix would be to insert MinVer into GenerateNuspecDependsOn. I just saw that it was done on GetPackageVersionDependsOn but it looks like it could be also added to GenerateNuspecDepends?

Hm, actually nope, that won't work. `_GetOutputItemsFromPack` is not part of `GenerateNuspecDepends` so it could still make the version calculation wrong.

adamralph:
@bording any opinions on this?

MinVer/build/MinVer.targets
```diff
@@ -11,7 +11,8 @@
     <NoWarn>$(NoWarn);NU5105</NoWarn>
   </PropertyGroup>
 
-  <Target Name="MinVer" BeforeTargets="BeforeCompile;GetAssemblyVersion;CoreCompile;GenerateNuspec" Condition="'$(DesignTimeBuild)' != 'true' AND '$(MinVerSkip)' != 'true'">
+  <!-- Use _GetOutputItemsFromPack instead of GenerateNuSpec to make sure that MinVer will run before calculating NuGet outputs -->
+  <Target Name="MinVer" BeforeTargets="BeforeCompile;GetAssemblyVersion;CoreCompile;_GetOutputItemsFromPack" Condition="'$(DesignTimeBuild)' != 'true' AND '$(MinVerSkip)' != 'true'">
```
adamralph, line 15: @xoofx I had a look through the structured logs and I believe I understand the problem now. Indeed, the first instances of the incorrect `...1.0.0.nupkg` in the examples you give in https://github.com/adamralph/minver/issues/675 are indeed in the `_GetOutputItemsFromPack` task.

Taking a closer look, I noticed that `_GetOutputItemsFromPack` is a dependency of `GenerateNuspec`, which also has `$(GenerateNuspecDependsOn)` as its _first_ dependency, so if we inject the `MinVer` target into that variable, I think we should be covered.

I also discovered that `_GetOutputItemsFromPack` is run when `Clean` is run, since `_CleanPackageFiles` runs after `Clean` and depends on `_GetOutputItemsFromPack`. At the moment, the packages never get cleaned when `dotnet clean` is called, because MinVer is never executed as part of the target graph, so the versions of the packages is not determined correctly.

I believe https://github.com/adamralph/minver/pull/681 fixes both problems.

What do you think?

adamralph: @xoofx I just remembered about your [previous comment](https://github.com/adamralph/minver/pull/676/files#issuecomment-1049502373):

> > Another fix would be to insert MinVer into GenerateNuspecDependsOn. I just saw that it was done on GetPackageVersionDependsOn but it looks like it could be also added to GenerateNuspecDepends?
> 
> Hm, actually nope, that won't work. `_GetOutputItemsFromPack` is not part of `GenerateNuspecDepends` so it could still make the version calculation wrong.

I think it's fine to use `$(GenerateNuspecDependsOn)` because `GenerateNuspec` looks like this:

```xml
  <Target Name="GenerateNuspec"
          Condition="'$(IsPackable)' == 'true'"
          Inputs="@(NuGetPackInput)" Outputs="@(NuGetPackOutput)"
          DependsOnTargets="$(GenerateNuspecDependsOn);_CalculateInputsOutputsForPack;_GetProjectReferenceVersions;_InitializeNuspecRepositoryInformationProperties">
```

As you can see it depends firstly on `$(GenerateNuspecDependsOn)`, and secondly on `_CalculateInputsOutputsForPack`.

In turn, `_CalculateInputsOutputsForPack` looks like this:

```xml
  <Target Name="_CalculateInputsOutputsForPack" DependsOnTargets="_GetOutputItemsFromPack">
```

It depends on `_GetOutputItemsFromPack`.

This means that if we inject `MinVer` into `$(GenerateNuspecDependsOn)` then we know `MinVer` will run _before_ `_GetOutputItemsFromPack`.

xoofx: Yep, agreed, #681 should fix the problem. I tested it locally and it works fine!

xoofx:
Closing as fixed by #681

PR #681, opened 2022-03-04
fix target injection for clean and pack

### LINKED ISSUE #675: Invalid order with NuGet pack leading to invalid outputs

adamralph:
relates to #682
relates to #683

### DIFF
--- a/MinVer/build/MinVer.targets
+++ b/MinVer/build/MinVer.targets
@@ -2,6 +2,7 @@
 
   <PropertyGroup>
     <MSBuildAllProjects Condition="'$(MSBuildAssemblyVersion)' == '' Or '$(MSBuildAssemblyVersion)' &lt; '16.0'">$(MSBuildAllProjects);$(MSBuildThisFileFullPath)</MSBuildAllProjects>
+    <GenerateNuspecDependsOn>$(GenerateNuspecDependsOn);MinVer</GenerateNuspecDependsOn>
     <GetPackageVersionDependsOn>$(GetPackageVersionDependsOn);MinVer</GetPackageVersionDependsOn>
   </PropertyGroup>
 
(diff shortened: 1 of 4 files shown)


adamralph:
I want to add tests for #682 and #683 before merging this.

bording:
I've been away from GitHub for the past two weeks, so I'm just catching up with things now.

Looking over the issues and this PR, adding the `MinVer` target to `GenerateNuspecDependsOn` seems like a good fix for #682.

For #683, I think we might want to tweak this a bit more, though.

The `_CleanPackageFiles` target has a `'$(GeneratePackageOnBuild)' == 'true'` condition, so it only runs when someone has set that to true, which is not the default.

The change here means that MinVer will be running during clean even when a package hasn't been built.

One option to fix this would be to add another target, something like:
```
<Target Name="_MinVerClean" BeforeTargets="Clean" DependsOnTargets="MinVer" Condition="'$(GeneratePackageOnBuild)' == 'true'" />
```

I believe that should continue to let the package be cleaned properly, but not invoke MinVer unnecessarily. It couldn't hurt to confirm that via a build log file, though.

adamralph:
> The change here means that MinVer will be running during clean even when a package hasn't been built.

@bording good catch! Your suggested change sounds good. I'll re-open https://github.com/adamralph/minver/issues/683 for now.

adamralph:
raised https://github.com/adamralph/minver/pull/704

PR #699, opened 2022-03-06
upgrade to net7.0

adamralph:


.github/workflows/ci.yml
```diff
@@ -17,9 +17,9 @@ jobs:
               os: macos-10.15
               build: ./build.sh
             tests:
-              framework: net6.0
-              sdk: "6.0.200"
-              sdk-major-minor: "6.0"
+              framework: net7.0
+              sdk: "7.0.100-preview.1.22110.4"
+              sdk-major-minor: "7.0"
```
adamralph, line 22: Still need to decide which SDK to use for older platforms. Maybe 6.0 (LTS) is more appropriate?

PR #725, opened 2022-04-03
add MinVerIgnoreLeadingZeros option

adamralph:
spun off from https://github.com/adamralph/minver/pull/720

### DIFF
--- a/MinVer.Lib/Version.cs
+++ b/MinVer.Lib/Version.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
 using System.Globalization;
 using System.Linq;
+using System.Text;
 using NuGet.Versioning;
 
 namespace MinVer.Lib;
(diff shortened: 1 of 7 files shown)


adamralph:
_continuing from https://github.com/adamralph/minver/pull/720#issuecomment-1086065049_

@bording if this is added, will you use it anywhere? For example, [the conversion of Particular repos to MinVer](https://github.com/Particular/Particular.Packaging/blob/29a1c8f6c4724fed10fae31302cd5ddde02fc0dd/src/Particular.Packaging/Particular.Packaging.csproj#L17)?

bording:
There was one repo where the most recent tag was in an invalid format that I had to manually fix. It could have been more widespread, but it was brought to the attention of the team doing the tagging that the format was invalid early enough to not be a larger problem.

If this feature existed, I probably would have just enabled it instead of going through the process of disabling the release workflow, deleting the bad tag, creating a valid tag, and then re-enabling the release workflow.

adamralph:
@bording given you've only found one problematic tag so far and you worked around it quite easily, I'm inclined to wait until we have more evidence this feature is required. In fact, it seems better to fix a single tag and _not_ use this feature, otherwise the floor is open for people to keep adding tags with leading zeros in the future, which doesn't seem to be a requirement. If, however, you come across a repo which has lots of tags with leading zeros then perhaps there's a use case for this feature, but again, I still wonder whether it's not better to do a one-off conversion of those tags and stick to properly formed tags in the future.

adamralph:
I'm going to close this PR for now, and move the branch to `archive/ignore-leading-zeros`. We can revisit as required.

PR #728, opened 2022-04-10
add release workflow

adamralph:


adamralph:
will open another PR without a `release-` name

PR #839, opened 2022-11-11
replace default pre-release phase with default pre-release identifiers

adamralph:
## Use case(s)

When the default pre-release identifiers, `alpha.0`, which are used to version post-RTM commits which are not tagged with a version, do not match the desired pre-release identifier conventions. For example, the desired conventions may not allow for `alpha.1` as the initial pre-release or they may have a different number of identifiers.

## Description

By setting `MinVerDefaultPreReleaseIdentifiers`, you can change the default pre-release identifiers from `alpha.0`, to something else, such as `preview.0`, or `preview.0.0`.

E.g.

```xml
<PropertyGroup>
  <MinVerDefaultPreReleaseIdentifiers>preview.0.0</MinVerDefaultPreReleaseIdentifiers>
</PropertyGroup>
```

## Alternatives

- Ignore, or workaround, the fact that the versions produced between an RTM tag and the next tag do match your desired pre-release identifier conventions.
- Tag the commit immediately following the RTM tag with a version with sentinel pre-release identifiers. For example, tag the commit immediately following the `1.0.0` RTM tag with `1.0.0-preview.0.0`.

## Additional context

See https://github.com/adamralph/minver/discussions/796#discussioncomment-4116854.


adamralph:
@iampluque @lonix1 with reference to https://github.com/adamralph/minver/discussions/796, I've released this in [4.3.0-beta.1](https://www.nuget.org/packages/MinVer/4.3.0-beta.1). Please feel free to give it a try.

PR #844, opened 2022-11-26
drop support for .NET Core

adamralph:
To safely and simply take advantage of new features.

I'm doing this now because the last version, .NET Core 3.1, [goes out of support](https://dotnet.microsoft.com/en-us/download/dotnet) on December 13, 2022.

### DIFF
--- a/.cspell.json
+++ b/.cspell.json
@@ -29,7 +29,6 @@
     "MINVERVERSIONOVERRIDE",
     "MyGet",
     "Nerdbank",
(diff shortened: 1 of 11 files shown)


adamralph:
released in 5.0.0-alpha.1 ([MinVer](https://www.nuget.org/packages/MinVer/5.0.0-alpha.1), [minver-cli](https://www.nuget.org/packages/minver-cli/5.0.0-alpha.1))

PR #868, opened 2023-04-12
Document shallow fetch behavior in Azure Pipelines

Crossbow78:
Follow-up from discussion post: https://github.com/adamralph/minver/discussions/865


adamralph:
Thanks @Crossbow78!

PR #894, opened 2023-06-11
include build metadata in informational verison only

### LINKED ISSUE #866: Do not include MinVerBuildMetadata in PackageVersion
## Use case(s)

It's useful to set `MinVerBuildMetadata` to get the build metadata into the [assembly informational version](https://learn.microsoft.com/en-us/dotnet/standard/assembly/versioning#assembly-informational-version). However, it is usually undesirable to have the build meta data added to the package version.

## Description

Currently, the value assigned to `MinVerBuildMetadata` is included in `MinVerVersion`, which is used to set `PackageVersion`. That means that the build metadata forms part of the package version:

Package file name: LibGit2Sharp.0.27.0.nupkg

image

With this change, the build metadata is only included in the [assembly informational version](https://learn.microsoft.com/en-us/dotnet/standard/assembly/versioning#assembly-informational-version).

## Alternatives

- A custom target to fix the `PackageVersion` property.
- Stop using `MinVerBuildMetadata` to get the value into the informational version, which is where I really want it

## Additional context

This is a problem because when you upload a package to nuget.org, the build metadata is included in the version. 

Sure, I can fix it with a custom target, but it's a pain to have do that. I can't see any real value in having the build metadata in there given how NuGet treats it, and I imagine that most people aren't going to want that in there, especially on a release package.

BTW, I hit this while in the process of releasing a new version of LibGit2Sharp. I just [switched it over to MinVer](https://github.com/libgit2/libgit2sharp/pull/2028) this weekend.

adamralph:
closes #866


MinVer/build/MinVer.targets
```diff
@@ -62,8 +62,10 @@
       <MinVerBuildMetadata Condition="$(MinVerVersion.Contains(`+`))">$(MinVerVersion.Split(`+`, 2)[1])</MinVerBuildMetadata>
       <AssemblyVersion>$(MinVerMajor).0.0.0</AssemblyVersion>
       <FileVersion>$(MinVerMajor).$(MinVerMinor).$(MinVerPatch).0</FileVersion>
-      <PackageVersion>$(MinVerVersion)</PackageVersion>
-      <Version>$(MinVerVersion)</Version>
+      <InformationalVersion>$(MinVerVersion)</InformationalVersion>
```
bording, line 65: The one thing that comes to mind about directly setting `InformationalVersion` here is how it interacts with SourceLink, which adds the commit to the `InformationalVersion`.

Seems like we should check to make sure that still works properly with this change.

adamralph: Good point. I’ll try to add a test for that.

adamralph: @bording I've added the test and it seems to work.

bording (approved):

PR #928, opened 2023-10-29
target .NET 8

adamralph:


adamralph:
The `lint` check fails due to .NET 7 being used. See https://github.com/super-linter/super-linter/pull/5047

PR #971, opened 2024-01-17
use a custom MSBuild task

### LINKED ISSUE #112: [2025-11-10] Use a custom MSBuild task

adamralph:
closes #112

### DIFF
--- a/MinVer/MinVer.csproj
+++ b/MinVer/MinVer.csproj
@@ -18,11 +18,9 @@
     <SuppressDependenciesWhenPacking>true</SuppressDependenciesWhenPacking>
     <TargetFrameworks>net6.0</TargetFrameworks>
     <LangVersion>default</LangVersion>
-    <Version>6.0.0</Version>
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.1.1" PrivateAssets="All" />
     <PackageReference Include="Microsoft.Build.Utilities.Core" Version="17.8.3" PrivateAssets="All" ExcludeAssets="runtime" />
   </ItemGroup>
 
--- a/MinVer/MinVer.csproj
+++ b/MinVer/MinVer.csproj
@@ -9,26 +9,21 @@
     <IncludeBuildOutput>false</IncludeBuildOutput>
     <!-- NU5100: The folder being packaged contains an assembly file that is not in lib folder. -->
     <NoWarn>$(NoWarn);NU5100</NoWarn>
-    <OutputType>Exe</OutputType>
     <PackageIcon>minver.png</PackageIcon>
     <PackageLicenseExpression>Apache-2.0</PackageLicenseExpression>
     <PackageProjectUrl>https://github.com/adamralph/minver</PackageProjectUrl>
     <PackageReadmeFile>README.md</PackageReadmeFile>
     <PackageReleaseNotes>https://github.com/adamralph/minver/blob/main/CHANGELOG.md</PackageReleaseNotes>
     <PublishRepositoryUrl>true</PublishRepositoryUrl>
-    <RollForward>major</RollForward>
     <SuppressDependenciesWhenPacking>true</SuppressDependenciesWhenPacking>
-    <TargetFrameworks>net6.0;net7.0;net8.0</TargetFrameworks>
+    <TargetFrameworks>net6.0</TargetFrameworks>
     <LangVersion>default</LangVersion>
+    <Version>6.0.0</Version>
   </PropertyGroup>
 
   <ItemGroup>
-    <Compile Include="..\minver-cli\Options.cs" Link="Options.cs" />
-    <Compile Include="..\minver-cli\Program.cs" />
-  </ItemGroup>
-
-  <ItemGroup>
-    <PackageReference Include="McMaster.Extensions.CommandLineUtils" Version="4.1.0" PrivateAssets="All" Publish="true" />
+    <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.1.1" PrivateAssets="All" />
+    <PackageReference Include="Microsoft.Build.Utilities.Core" Version="17.8.3" PrivateAssets="All" ExcludeAssets="runtime" />
   </ItemGroup>
 
   <ItemGroup>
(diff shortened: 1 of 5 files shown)


adamralph:
@bording I cherry-picked this from https://github.com/bording/minver/commits/msbuild-task-spike/

MinVer/MinVer.csproj
```diff
@@ -9,26 +9,21 @@
...
     <PackageProjectUrl>https://github.com/adamralph/minver</PackageProjectUrl>
     <PackageReadmeFile>README.md</PackageReadmeFile>
     <PackageReleaseNotes>https://github.com/adamralph/minver/blob/main/CHANGELOG.md</PackageReleaseNotes>
     <PublishRepositoryUrl>true</PublishRepositoryUrl>
-    <RollForward>major</RollForward>
     <SuppressDependenciesWhenPacking>true</SuppressDependenciesWhenPacking>
-    <TargetFrameworks>net6.0;net7.0;net8.0</TargetFrameworks>
+    <TargetFrameworks>net6.0</TargetFrameworks>
     <LangVersion>default</LangVersion>
+    <Version>6.0.0</Version>
```
adamralph, line 21: @bording why is this here?

MinVer/MinVer.csproj
```diff
@@ -9,26 +9,21 @@
...
   </PropertyGroup>
 
   <ItemGroup>
-    <Compile Include="..\minver-cli\Options.cs" Link="Options.cs" />
-    <Compile Include="..\minver-cli\Program.cs" />
-  </ItemGroup>
-
-  <ItemGroup>
-    <PackageReference Include="McMaster.Extensions.CommandLineUtils" Version="4.1.0" PrivateAssets="All" Publish="true" />
+    <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.1.1" PrivateAssets="All" />
```
adamralph, line 25: @bording are you sure we need this? I tried building without it and the `AssemblyInformationalVersion` had the commit SHA added, so I think just using the .NET 8 SDK to build the project is enough, even though the project targets `net6.0`.

MinVer/build/MinVer.targets
```diff
@@ -4,8 +4,12 @@
     <MSBuildAllProjects Condition="'$(MSBuildAssemblyVersion)' == '' Or '$(MSBuildAssemblyVersion)' &lt; '16.0'">$(MSBuildAllProjects);$(MSBuildThisFileFullPath)</MSBuildAllProjects>
     <GenerateNuspecDependsOn>$(GenerateNuspecDependsOn);MinVer</GenerateNuspecDependsOn>
     <GetPackageVersionDependsOn>$(GetPackageVersionDependsOn);MinVer</GetPackageVersionDependsOn>
+    <MinVerTaskPath Condition="'$(MSBuildRuntimeType)' == 'Core'">$(MSBuildThisFileDirectory)..\minver\net6.0\MinVer.dll</MinVerTaskPath>
+    <MinVerTaskPath Condition="'$(MSBuildRuntimeType)' != 'Core'">$(MSBuildThisFileDirectory)..\minver\net472\MinVer.dll</MinVerTaskPath>
```
adamralph, line 8: @bording this doesn't seem right. The `MinVer` project isn't targetting `net472`.

MinVer/MinVer.csproj
bording: I just wanted to increase the version in my spike to make it easier to try out.

MinVer/MinVer.csproj
bording: At the time of the spike, you had not moved the repo over to the .NET 8 SDK. You're correct that you don't need it anymore.

MinVer/build/MinVer.targets
bording: > The MinVer project isn't targetting net472.

It will need to. I didn't bother with changing that in my spike, but you'll need to multi-target and build a `net472` version of the task assembly as well to support .NET Framework MSBuild, which is used by Visual Studio.

MinVer/MinVer.csproj
adamralph: Ah OK, I misunderstood the effect of the cherry-pick. I thought the commit in your spike branch was adding this line.

MinVer/build/MinVer.targets
bording: yep

adamralph: Hmmm... what concerns me about this is that the `net472` code path will be untested in CI. Local smoke testing would be one option to mitigate that but it's not easy for me to do so since I no longer use Windows. I guess it's unlikely that the `net472` version will have problems if the `net6` version doesn't but it's the edge cases/unknown unknowns that I'm wary of. I have to say this gives me cold feet about switching to an MSBuild task.

Is the caching part of your spike only possible if we switch to an MSBuild task?

bording: > Hmmm... what concerns me about this is that the net472 code path will be untested in CI

That doesn't have to be true, though. It would certainly be possible to have tests that also verify behavior on `net472`.

> I have to say this gives me cold feet about switching to an MSBuild task.

I don't really understand this, and given the benefits of doing it, I really don't agree.

> Is the caching part of your spike only possible if we switch to an MSBuild task?

Yes, it's relying on APIs that only accessible from a task.

adamralph: > That doesn't have to be true, though. It would certainly be possible to have tests that also verify behavior on `net472`.

I guess it is possible, but it would not require running tests conditionally, on Windows only, calling `msbuild.exe`?

bording: I'm not sure what sort of tests you're thinking of. Most of the logic would just be testable directly, and the test projects could be multi-targeted, so that would take care of that.

If you want some sort of integration test, then that would need to use `dotnet`/`msbuild.exe` and to use the respective task assemblies, but that also seems rather doable?

adamralph:
closing in favour of https://github.com/adamralph/minver/pull/989

PR #994, opened 2024-04-06
WIP - Exploration to expose extended git details to MSBuild

robmen:
@adamralph I'm opening this PR to see if you are interested in changes like these. It exposes more git information to MSBuild that allows us to use the MinVer concepts in our more complicated build systems like in the WiX Toolset. 

If you are interested in these additions, I'll need to clean up the code quite a bit, and I would appreciate any feedback you have at this time so that I can factor it to your liking.

If you are NOT interested in the complexity added by this PR, that's totally cool. I really appreciate the ideas you put forward in MinVer. It changed (simplified!) the way I think about versioning using git.

There are two "features" in here. I'm sorry. I should have separated them into separate commits, and I will do that if this PR interests you.

1. Expose the various pieces of the git version to MSBuild. This allows us to do various string concatenations and (in some cases) math operations in MSBuild .targets files to address the various versions in our build system.

2. Adds a boolean to include the default pre-release labels even when the tag has pre-release information. We use default pre-release labels like `dev` on all dev builds. After releasing an `rc-1` we like the dev builds to still have the `dev` label, like `1.0.0-rc.1.dev.1`. This is mostly a vanity feature, but it turned out to be pretty straightforward to wire in so I took a shot at it. If you like this feature, I'd also add it to the minver.exe to maintain consistency.

Let me know if you are interested. Again, I know the code needs to be cleaned up but I didn't want to do too much if it turned out this added more complexity than you want in MinVer (that's why I jokingly called the new MSBuild task `SomeVer` 😆).

### DIFF
--- a/MinVer.Lib/Git.cs
+++ b/MinVer.Lib/Git.cs
@@ -8,6 +8,7 @@ namespace MinVer.Lib;
 internal static class Git
 {
     private static readonly char[] newLineChars = ['\r', '\n',];
+    private static readonly char[] spaceChars = [' '];
 
     public static bool IsWorkingDirectory(string directory, ILogger log) => GitCommand.TryRun("status --short", directory, log, out _);
 
@@ -30,7 +31,7 @@ internal static class Git
         var commits = new Dictionary<string, Commit>();
 
         foreach (var shas in lines
-            .Select(line => line.Split(" ", StringSplitOptions.RemoveEmptyEntries)))
+            .Select(line => line.Split(spaceChars, StringSplitOptions.RemoveEmptyEntries)))
         {
             commits.GetOrAdd(shas[0], () => new Commit(shas[0]))
                 .Parents.AddRange(shas.Skip(1).Select(parentSha => commits.GetOrAdd(parentSha, () => new Commit(parentSha))));
@@ -45,7 +46,7 @@ internal static class Git
         GitCommand.TryRun("show-ref --tags --dereference", directory, log, out var output)
             ? output
                 .Split(newLineChars, StringSplitOptions.RemoveEmptyEntries)
-                .Select(line => line.Split(" ", 2))
+                .Select(line => line.Split(spaceChars, 2))
                 .Select(tokens => (tokens[1][10..].RemoveFromEnd("^{}"), tokens[0]))
             : Enumerable.Empty<(string, string)>();
 
--- a/MinVer.Lib/MinVer.Lib.csproj
+++ b/MinVer.Lib/MinVer.Lib.csproj
@@ -2,12 +2,13 @@
 
   <PropertyGroup>
     <IsPackable>false</IsPackable>
-    <TargetFrameworks>net6.0;net7.0;net8.0</TargetFrameworks>
+    <TargetFrameworks>net472;net6.0;net7.0;net8.0</TargetFrameworks>
     <LangVersion>default</LangVersion>
   </PropertyGroup>
 
   <ItemGroup>
     <PackageReference Include="NuGet.Versioning" Version="6.8.0" />
+    <PackageReference Include="PolySharp" Version="1.14.1" PrivateAssets="all" IncludeAssets="runtime; build; native; contentfiles; analyzers" />
   </ItemGroup>
 
 </Project>
--- a/MinVer.Lib/Version.cs
+++ b/MinVer.Lib/Version.cs
@@ -40,16 +40,26 @@ public class Version : SemanticVersion
             : new Version(minMajorMinor.Major, minMajorMinor.Minor, 0, defaultPreReleaseIdentifiers.ToList(), this.height, this.Metadata);
     }
 
-    public Version WithHeight(int newHeight, VersionPart autoIncrement, IEnumerable<string> defaultPreReleaseIdentifiers) =>
-        this.preReleaseIdentifiers.Count == 0 && newHeight > 0
-            ? autoIncrement switch
+    public Version WithHeight(int newHeight, VersionPart autoIncrement, IEnumerable<string> defaultPreReleaseIdentifiers, bool includeDefaultPreReleaseIdentifiersWithPrereleases)
+    {
+        if (this.preReleaseIdentifiers.Count == 0 && newHeight > 0)
+        {
+            return autoIncrement switch
             {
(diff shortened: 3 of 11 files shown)


adamralph:
@robmen

> 1. Expose the various pieces of the git version to MSBuild. This allows us to do various string concatenations and (in some cases) math operations in MSBuild .targets files to address the various versions in our build system.

As stated in the [README](https://github.com/adamralph/minver?tab=readme-ov-file#version-numbers), MinVer already sets properties representing every part of the version. What more are you proposing?

> 2. Adds a boolean to include the default pre-release labels even when the tag has pre-release information. We use default pre-release labels like `dev` on all dev builds. After releasing an `rc-1` we like the dev builds to still have the `dev` label, like `1.0.0-rc.1.dev.1`. This is mostly a vanity feature, but it turned out to be pretty straightforward to wire in so I took a shot at it. If you like this feature, I'd also add it to the minver.exe to maintain consistency.

I'm not sure I understand the use case.

You tag and release `1.0.0-rc.1`. You then push some commits to that branch which result in `1.0.0-rc.1.1`, `1.0.0-rc.1.2`, etc. At some point you tag and release `1.0.0-rc.2`. Why does that not work for you?

robmen:
1. Here's the full list of properties exposed:

   * `Version`
   * `Major`
   * `Minor`
   * `Patch`
   * `Label`
   * `Sha` - new
   * `ShortSha` - new
   * `Height` - new

2. It _could_ work. Today we use labels akin to `-rc.1.dev.#`, and it was straightforward to enhance MinVer to continue to support it.

Let me know if there are any other questions I can answer.

adamralph:
> Here's the full list of properties exposed:
> 
> * `Version`
> * `Major`
> * `Minor`
> * `Patch`
> * `Label`
> * `Sha` - new
> * `ShortSha` - new
> * `Height` - new

What is your use case for the SHA?

I would avoid using height in isolation. It's part of the pre-release identifiers, and they should really be treated as a discrete item and not broken apart. But I guess this is driven by your desire to insert `dev` before the height, so we should probably examine that part first.

> 2\. It _could_ work. Today we use labels akin to `-rc.1.dev.#`, and it was straightforward to enhance MinVer to continue to support it.

What purpose does the `dev` serve? Is it enough to know that any pre-release version with an ordinal of `x.y` (e.g. `rc.1.1`) is a "dev" version? If you really do need `dev` as another indicator, I would suggest tagging it on at the end:

```xml
<Target Name="MyTarget" AfterTargets="MinVer" Condition="'$([System.String]::new($(MinVerPreRelease)).Split(`.`).Length)' == '3'" >
  <PropertyGroup>
    <PackageVersion>$(MinVerVersion).dev</PackageVersion>
    <Version>$(PackageVersion)</Version>
  </PropertyGroup>
</Target>
```

adamralph:
BTW, in .NET 8 SDK, it seems that [`SourceRevisionId` will give you the SHA](https://learn.microsoft.com/en-us/dotnet/core/compatibility/sdk/8.0/source-link#new-behavior).

robmen:
Ahh, sorry, I forgot to motivate the change. There are three goals in play.

1. The primary goal is to enable old-school four-part file versions. The height is used as the fourth digit in: `{Major}.{Minor}.{Patch}.{Height}`. That allows you to enable build-to-build upgrades when not using SemVer. However, I realize I've made a mistake and need the "full height" (the height since the last non-prerelease tag), not the height since the last tag. I'll need to fix that and probably rename the `Height` output property `FullHeight` (or a better name).

2. The goal for the SHA is to differentiate builds from different developers that may have the same height because they are working in private topic branches. The `SourceRevisionId` may work; I'd have to create a .NET 8 .csproj to verify. But MinVer knows the exact SHA for the candidate, so it was safest to return that directly.

3. The final goal is to enable our current way of differentiating "official builds" (releases without `dev.` label) from "dev builds" (releases with `dev.` label). Without the `dev.` feature, the first range of dev builds would be `1.0.0-dev.#`. Then, after the first RC, dev builds would end in `.#` like `1.0.0-rc1.#` and so on. It's easier to communicate with consumers that a dev build is a build with `dev.` in it.

Hopefully, I'm motivating the scenarios better. I will need to make that fix for full height, but I don't want to waste your time if you are not interested in adding any of these changes to MinVer. :)

adamralph:
> 1. The primary goal is to enable old-school four-part file versions. The height is used as the fourth digit in: `{Major}.{Minor}.{Patch}.{Height}`. That allows you to enable build-to-build upgrades when not using SemVer. However, I realize I've made a mistake and need the "full height" (the height since the last non-prerelease tag), not the height since the last tag. I'll need to fix that and probably rename the `Height` output property `FullHeight` (or a better name).

I would advise against using the height for this. As stated in the [FAQ](https://github.com/adamralph/minver?tab=readme-ov-file#can-i-use-the-version-calculated-by-minver-for-other-purposes), the [official guidance for assembly file version](https://learn.microsoft.com/en-ca/dotnet/standard/library-guidance/versioning#assembly-file-version) states:

> ✔️ CONSIDER including a continuous integration build number as the AssemblyFileVersion revision.

The FAQ shows how to do that.

---

> 2\. The goal for the SHA is to differentiate builds from different developers that may have the same height because they are working in private topic branches. The `SourceRevisionId` may work; I'd have to create a .NET 8 .csproj to verify. But MinVer knows the exact SHA for the candidate, so it was safest to return that directly.

`SourceRevisionId` should work for this. AFAIK you don't have to be _targetting_ .NET 8, you just have to be using the .NET 8 SDK to build the project.

---

> 3\. The final goal is to enable our current way of differentiating "official builds" (releases without `dev.` label) from "dev builds" (releases with `dev.` label). Without the `dev.` feature, the first range of dev builds would be `1.0.0-dev.#`. Then, after the first RC, dev builds would end in `.#` like `1.0.0-rc1.#` and so on. It's easier to communicate with consumers that a dev build is a build with `dev.` in it (e.g. `1.0.0-dev.#` and `1.0.0-rc1.dev.#`).

I would do this as I suggested in my previous answer, which would give you versions like `1.0.0-rc.1.1.dev`. This requires no change to MinVer.

robmen:
Alright, cool. Thank you very much for the insights.

PR #1020, opened 2024-07-20
move build binaries from minver to build/bin in nupkg

adamralph:


adamralph:
released in [5.1.0-rc.1](https://www.nuget.org/packages/MinVer/5.1.0-rc.1)

adamralph:
re-released as [6.0.0-rc.1](https://www.nuget.org/packages/MinVer/6.0.0-rc.1) because the next release contains at least one breaking change

PR #1021, opened 2024-07-20
MSBuild caching

adamralph:


adamralph:
@bording what do you think of this as a lightweight alternative to #989?

I was thinking it would be nice if MSBuild exposed tasks which could read and write from the build cache, so I went ahead and wrote those tasks. I could even see those potentially being contributed to MSBuild.

MSBuild.Caching/MSBuild.Caching.csproj
```diff
@@ -0,0 +1,13 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <IsPackable>false</IsPackable>
+    <TargetFrameworks>net472;net6.0</TargetFrameworks>
+    <LangVersion>12.0</LangVersion>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Build.Utilities.Core" Version="17.0.1" PrivateAssets="All" ExcludeAssets="runtime" />
```
adamralph, line 10: @bording I cherry picked this from https://github.com/adamralph/minver/pull/989

What determines the version of this package that should be referenced?

MSBuild.Caching/CacheSet.cs
```diff
@@ -0,0 +1,19 @@
...
+public class CacheSet : Task
+{
+    public string? Key { get; set; }
+
+    public string? Value { get; set; }
+
+    public override bool Execute()
+    {
+        this.BuildEngine4.RegisterTaskObject(
+            this.Key, this.Value, RegisteredTaskObjectLifetime.Build, allowEarlyCollection: false);
```
adamralph, line 15: Since it's only a version string, I thought to switch off early collection, as the cost of running MinVer again seems more than keeping the string around longer.

From https://learn.microsoft.com/en-us/dotnet/api/microsoft.build.framework.ibuildengine4.registertaskobject (emphasis mine) —

> It is strongly recommended that allowEarlyCollection be set to true _if the object will retain any significant amount of data…_

adamralph:
FWIW, I tested this with https://github.com/Particular/ServiceControl, and a single threaded build went from 43 seconds to 24 seconds.

adamralph:
testing the water with https://github.com/dotnet/msbuild/pull/10418

MSBuild.Caching/CacheSet.cs
rainersigwald: MSBuild doesn't actually pay attention to this any more:

https://github.com/dotnet/msbuild/blob/8b03742aafd18a26a5d03cd045e4d24963886503/src/Shared/RegisteredTaskObjectCacheBase.cs#L59-L64

(We should update the docs though: https://github.com/dotnet/msbuild/issues/10425

adamralph: Ah, thanks for the info!

MSBuild.Caching/MSBuild.Caching.csproj
bording: My understanding is that the package version corresponds to the MSBuild version, so you would want to reference the version of the package that corresponds to the version of MSBuild that shipped in the oldest .NET SDK you're trying to support.

Since this is targeting `net6.0`, that would correspond to the version of MSBuild that ships in the 6.0.100 SDK.

bording (approved):
Overall I still prefer moving all the logic to an MSBuild task instead of calling other processes during the build, but this seems like an okay intermediate step.

MSBuild.Caching/MSBuild.Caching.csproj
adamralph: Thanks. On that assumption, as far as I can tell (it isn't easy), 17.0.1 is the right version.

adamralph:
@bording I've released this in [5.1.0-rc.1](https://www.nuget.org/packages/MinVer/5.1.0-rc.1)—please feel free to give it a try.

MSBuild.Caching/MSBuild.Caching.csproj
baronfel: In the future you can reference this chart we made to help answer this question: https://aka.ms/dotnet/matrixofpaine

adamralph: @baronfel that's very useful, thanks!

adamralph: @baronfel so useful in fact, that I decided to bake it into the source — https://github.com/adamralph/minver/pull/1047/files

PR #1031, opened 2024-07-21
tagged version wins even if MinVerMinimumMajorMinor is higher

### LINKED ISSUE #999: Tagged version wins even if MinVerMinimumMajorMinor is higher
## Use case(s)

`MinVerMinimumMajorMinor` is provided as way to say: future dev work on this branch is for version x.y, rather than patches or minors on the previous version. It seems reasonable for someone to later decide that a commit on the branch is actually a patch or minor on the previous version, and tags the commit as such and expects that version to pop out.

## Description

As currently [stated by the docs](https://github.com/adamralph/minver?tab=readme-ov-file#how-it-works), "If the current commit has a version tag…The version is used as-is", without any mention of the `MinVerMinimumMajorMinor`. Described as a bug in https://github.com/adamralph/minver/issues/996 but this issue reframes the problem as a (breaking) enhancement.

## Alternatives

Leave it as is and change the docs to describe the actual behavior.

## Additional context

N/A

adamralph:
closes #999

### DIFF
--- a/MinVer.Lib/Versioner.cs
+++ b/MinVer.Lib/Versioner.cs
@@ -10,25 +10,32 @@ public static class Versioner
 
         var defaultPreReleaseIdentifiersList = defaultPreReleaseIdentifiers.ToList();
 
-        var (version, height) = GetVersion(workDir, tagPrefix, defaultPreReleaseIdentifiersList, log);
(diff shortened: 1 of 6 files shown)


adamralph:
@DavidBoike what do you think?

DavidBoike (approved):

